PR #15, opened 2017-01-24
Updated project to allow support for net452 framework target

jorjenava:
My team likes workflow core and where it's headed but we unfortunately still need to use the full .net framework.  It looks like the code all still works when targeting net452 so I thought I'd do a pull request and maybe help some others out.


danielgerlag:
Thank you

PR #35, opened 2017-05-23
Lock each list of memory persistence before using

qazq:
In some cases, add/remove workflow at persistence store will throw an
exception. ("Collection was modified; enumeration operation may not execute.")

### DIFF
--- a/src/WorkflowCore/Services/MemoryPersistenceProvider.cs
+++ b/src/WorkflowCore/Services/MemoryPersistenceProvider.cs
@@ -22,66 +22,90 @@ namespace WorkflowCore.Services
 
         public async Task<string> CreateNewWorkflow(WorkflowInstance workflow)
         {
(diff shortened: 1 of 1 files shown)


danielgerlag:
Which persistence provider are you using?

qazq:
MemoryPersistenceProvider

qazq:
This is my testing code.
```cs
            var host = serviceProvider.GetService<IWorkflowHost>();
            host.RegisterWorkflow<HelloWorldWorkflow>();        
            host.Start();

            Thread.Sleep(100);
            for (int i = 0; i < 100; i++)
            {
                host.StartWorkflow("HelloWorld", 1, null);
                if (i % 10 == 0)
                {
                    Thread.Sleep(1000);
                }
            }
```

It has the chance to happen the exception.

qazq:
ConcurrentBag<> seems cannot remove a specific item, so I add the locks.

PR #40, opened 2017-06-05
[UnitTest] Add ConcurrentPersistWorkflow() test item

qazq:
To check the persistence provider if occurs resource competition.

### DIFF
--- a/test/WorkflowCore.UnitTests/BasePersistenceFixture.cs
+++ b/test/WorkflowCore.UnitTests/BasePersistenceFixture.cs
@@ -6,6 +6,7 @@ using WorkflowCore.Models;
 using Xunit;
 using FluentAssertions;
 using WorkflowCore.TestAssets;
+using System.Threading.Tasks;
 
 namespace WorkflowCore.UnitTests
 {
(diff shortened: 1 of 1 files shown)


qazq:
It's strange. I create 30 workflow instances by the Task<>. The MemoryPersistenceProvider takes about 80ms, but the SqlitePersistenceProvider takes over 5 seconds. If I increase the number of workflow instances to 100, MemoryPersistenceProvider takes 84ms and SqlitePersistenceProvider takes 15 seconds. The SQLite's performance should better than it. I don't have PostgreSQL server now, so I didn't test it.

danielgerlag:
@qazq it could be the cost of entity framework setting up the schema... I will try look at this on the weekend

test/WorkflowCore.UnitTests/BasePersistenceFixture.cs
```diff
@@ -92,5 +93,58 @@ public void PersistWorkflow()
...
+                    }
+                }));
+            }
+
+            Task.WhenAll(tasks).Wait();
+
+            foreach (var task in tasks)
+            {
+                task.Result.Should().NotBeNull();
+            }
```
danielgerlag: Awesome... instead of catching the exception and returning null... could we perhaps let the exception bubble up and assert it did not happen with fluent assertions?
http://fluentassertions.com/documentation.html#exceptions

qazq: I have tried it, but `ShouldNotThrow()` cannot be used by Task (need Action). I need invoke many tasks to access persist provider concurrently. `ShouldNotThrow()` is a blocking call, so I cannot do them in concurrently. Do you have any idea?

danielgerlag: You could try creating a list of `Action` and then using `Parallel.Invoke(actions)` or `Parallel.ForEach` from `System.Threading.Tasks`?

qazq: Thank you. I've modified.

PR #41, opened 2017-06-13
[UnitTest] Add SqlServer testing

qazq:


qazq:
I haven't used Docker. I tried it, but I got a failed when I setting the environment. Could you help me to modify the code to support Docker? Thank you.

PR #68, opened 2017-10-26
Support asp.net core 2.0

qazq:
In my project, I have to upgrade to EF core 2.0 to fix some critical bugs. But workflow-core is not compatible with EF Core 2.0. I update all the workflow-core projects to support .net core 2.0


src/WorkflowCore/WorkflowCore.csproj
```diff
@@ -25,14 +24,15 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="1.1.1" />
-    <PackageReference Include="Microsoft.Extensions.Logging" Version="1.1.1" />
-    <PackageReference Include="Microsoft.Extensions.ObjectPool" Version="1.1.1" />
+    <PackageReference Include="Microsoft.AspNetCore.Http.Abstractions" Version="2.0.0" />
```
danielgerlag: why do we need `Microsoft.AspNetCore.Http.Abstractions`?

qazq: Because I have to use EF core 2.0 to fix some critical bugs that came from EF core 1.x. My project already updated to 2.0 and seems conflicted with `WorkflowCore.Persistence.EntityFramework 1.x` so my project cannot work normally.

This PR is not required. If one day it needs to update to 2.0, maybe this can make it easier :)

BTW, `Microsoft.AspNetCore.Http.Abstractions` is my mistake.

harmandeol:
we are also facing same issue. is there any plan to upgrade soon ?

danielgerlag:
EF packages have been upgraded to EF Core 2.0

PR #108, opened 2018-02-22
Fix null complete time

qazq:


qazq:
Do you mean increment the version to 1.6.1

Issue #101, opened 2018-01-29
define UserTask using JSON file


kbe69:
Hi,

First of all, thanks for the great job.

I want to know if there is a way to define a workflow with usertask using JSON file?

Thanks

danielgerlag:
Yes, it should just be a matter of reference the correct class for the workflow step, as per these:
https://github.com/danielgerlag/workflow-core/tree/master/src/extensions/WorkflowCore.Users/Primitives

I'm heading out of town, but when I get back I will post some examples.

fubar-coder:
I'm not sure that this works, because the definition loader creates a step of type `WorkflowStep<TStepBody>` (with `TStepBody` being `UserTask`), but it should be `UserTaskStep` instead, because the two additional properties `Options` and `Escalations` seem to be required for the instantiation of `UserTask`.

The problem is two-fold:

1. Check if the `StepType` in `DefinitionLoader.ConvertSteps` is derived from `WorkflowStep` and use it as a step.
2. Find a generic way to fill step-defined properties that are set during the instantiation of the `TStepBody` (using the `ConstructBody` function).

I'll try to create a PR to fix the problem.

danielgerlag:
Thanks for the feedback... I think the best solution would be to refactor the `WorkflowCore.Users` package and remove the `UserTaskStep` class entirely by moving it's properties into `UserTask`, it then becomes possible to map values to those properties using the standard `Input` and `Output` methods.  I have already done this with the core primitives in the `WorkflowCore` package.

federicorOT:
Hi! @danielgerlag I'm loving the framework :D Is there any new about this? Can a User task be defined in a json file?

federicorOT:
Sorry @danielgerlag but do you have any example of this please? I still can not make it work :(
I'm not being able to create a user task through a json file

Tailslide:
> I think the best solution would be to refactor the WorkflowCore.Users package  ... I have already done this with the core primitives in the `WorkflowCore` package.

I'm also interested in this @danielgerlag but I'm not familiar enough with the project to implement the change. Any more guidance on what steps I could take in refactoring would be helpful.

federicorOT:
@Tailslide Hi, I realized this creating a new customUserTaskStep, and using Options and Escalations as Inputs

Tailslide:
@federicorOT Thanks for your help.  Can you please post some sample code?  I don't really understand your explanation. Are you subclassing UserTaskStep? How are you linking it to UserTask?

PR #113, opened 2018-03-12
Enable definition of user tasks using JSON file

fubar-coder:
Fixes issue #101.

CancelCondition and Saga aren't supported for this use case. The DefinitionLoader requires a new IServiceProvider parameter to enable instantiation of the workflow step using DI.


fubar-coder:
Some Codacy "issues" are plain wrong. It complains about making some functions `static`, but they cannot be made static because they require access to instance variables 😠 .

danielgerlag:
Don't worry about it 😄

PR #142, opened 2018-06-01
Add Castle.Windsor extensions to WorkflowCore.

st2forget:
As I realized that this project can also work with applications and almost all versions of .net, more over, it should supports another IOC as well, I did a little modification that:
1. Makes some properties/classes PUBLIC instead of internal so that it could be a good workaround for registering IOC
2. Add Castle.Windsor extensions which allows to AddWorkflow as normal and replace default ServiceProvider to Windsor's ServiceProvider


danielgerlag:
@Miggleness have a look at this article
https://kristian.hellang.com/third-party-dependency-injection-in-asp-net-core/

Issue #153, opened 2018-07-10
Assign workflow instance name/identifier


Miggleness:
I’ve a need to explicitly set an IT Ops friendly identifier for my workflow instances for searching and investigatory purposes. 

Is this supposed to be the purpose of the Workflow Instance Reference property?  If so, we should be able to assign a value to it when starting a new workflow. 

If that isnt the case we can come up with a solution and i can submit a PR

danielgerlag:
Yes, that was the intention of the `Reference` field, however, I never got around to exposing it anywhere.
A PR would be most welcome 😃

PR #156, opened 2018-07-10
Assign workflow instance name/identifier

Miggleness:
This PR is to address #153 which exposes the WorkflowInstance's 'Reference' property.

Notes:
* A number of unit tests are failing from my end prior to change, I didn't bother with those.  I'm using Rider 2018.1 on Mac.
* Regarding placement of "Reference" argument in  StartWorkflow().  The current placement feels awkward, but it sets up for slightly better compatibility. Preferably, the sequence will be (id, version, reference, data)
* I'm considering updating model builder to index 'Reference'.  But then, probably better to let the end users decide on that.

### DIFF
--- a/src/samples/WorkflowCore.Sample02/Program.cs
+++ b/src/samples/WorkflowCore.Sample02/Program.cs
@@ -21,7 +21,7 @@ namespace WorkflowCore.Sample02
             host.RegisterWorkflow<SimpleDecisionWorkflow>();
             host.Start();
 
-            host.StartWorkflow("Simple Decision Workflow", 1, null, null);
+            host.StartWorkflow("Simple Decision Workflow");
 
             Console.ReadLine();
             host.Stop();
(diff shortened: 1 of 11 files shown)


src/samples/WorkflowCore.Sample02/Program.cs
```diff
@@ -21,7 +21,7 @@ public static void Main(string[] args)
             host.RegisterWorkflow<SimpleDecisionWorkflow>();
             host.Start();
 
-            host.StartWorkflow("Simple Decision Workflow", 1, null);
+            host.StartWorkflow("Simple Decision Workflow", 1, null, null);
```
danielgerlag: Could we simplify all the samples to just use `host.StartWorkflow(name)`?

Miggleness:
done :)

Any chance you're upgrading the samples to dotnet core 2.1.1? For some odd reason, my Mac won't run the 1.x SDK installers.

PR #160, opened 2018-07-21
Set IServiceCollection as return type for AddWorkFlow

Kahbazi:


Kahbazi:
All of the codes in Microsoft.Extesions and almost all other libraries that are adding Services to `IServiceCollection` use extension method which returns `IServiceCollection`. The benefit is that we can use method chaining like below

```
new ServiceCollection()
    .AddWorkflow()
    .AddLogging()
    .BuildServiceProvider();
```

PR #161, opened 2018-07-22
Remove reflection for creating Data

Kahbazi:
Since Data for Workflow must have a parameterless constructor, this constraint can be added to `StartWorkflow<TData>`,  so we don't use reflection for creating `Data`


danielgerlag:
@Kahbazi I merged another PR just before this one, which is causing some conflicts

Kahbazi:
@danielgerlag I resolve the conflicts

PR #169, opened 2018-08-05
Support nested property for input and output of step

Kahbazi:
I changed `WorkflowExecutor` in order to support setting values for nested property


danielgerlag:
#245

Issue #95, opened 2017-12-28
Create dynamic DataModels composed by workflow def.steps


linkev44:
When i use JSON for creating a dynamic WF, i need to bind a C# strong typed model. We need a dynamic data model definition .. each step in JSON Workflow should create a part of this global model ... how you think that can be implemented ?

Sorry for my approximative english ..

danielgerlag:
Hi,

I'm not sure I understand what you are asking... could you try explain it in different terms?

linkev44:
Hi, 
When i will design a new workflow (by JSON) the workflow can be modified by users .. if an user add a step in dynamic workflow..  the workflow data model must be able to support this add of properties (needed by the new step) dynamically, like a property bag ... Do you see what i say ?

ewyuen:
Do you mean when defining a workflow with IWorkflow<TData>, you want TData to be a dynamic class or a class which can be dynamically added properties to it?  Do you consider just add a Dictionary<string, object> property to your TData class?

danielgerlag:
Could you use [System.Dynamic.ExpandoObject](https://msdn.microsoft.com/en-us/library/system.dynamic.expandoobject(v=vs.110).aspx) ?

urdeveloper:
Hi Daniel, is it possible to use ExpandoObject?I tried it before with JSON definition and didn't seem to work. Let's say you'd like to assign a step's output value to Data.X (here Data is type of ExpandoObject and X is not defined yet), it will error out that X is not a member of ExpandoObject.

danielgerlag:
I think there may be an issue with using dynamic types. Let me look into this further and get back to you.

danielgerlag:
I can confirm that it is not possible to use dynamic types with the workflow data object on the current version.  I will look into adding this functionality in the next release.

VadimLopatkin:
@danielgerlag hi, we are considering using your library in our app, but we really need the possibility to define workflows in JSON at runtime, including specifying the input and output variables. As a workaround I was thinking of changing the `WorkflowCore.Services.WorkflowExecutor.ProcessOutputs`  and `WorkflowCore.Services.WorkflowExecutor.ProcessInputs` such that they would work with `Dictionary<string, object>` (and issue a PR afterwards).

What do you think about it? Do you see a way the issue can be fixed?

danielgerlag:
It is already possible to define JSON workflows at runtime. How would thr suggested change make a difference?

VadimLopatkin:
[...]

PR #173, opened 2018-08-22
Added support for indexers instead of properties for the DataType.

dr-BEat:
I recently discovered your library and liked it quite a lot.
However I had a similar use case to issue #95, basically requiring a dynamic data model.

I decided to give a try to extending support for this without breaking old code or changing public interfaces.

My changes would allow any type with an indexer with a string parameter to be used.
You could even mix normal properties and "dynamic" ones.

This code is of course not perfect, so please let me know if you think this would be useful for workflow-core and if you have ideas for improvements.

### DIFF
--- a/src/WorkflowCore/Services/DefinitionStorage/DefinitionLoader.cs
+++ b/src/WorkflowCore/Services/DefinitionStorage/DefinitionLoader.cs
@@ -173,12 +173,15 @@ namespace WorkflowCore.Services.DefinitionStorage
                 var dataParameter = Expression.Parameter(dataType, "data");
                 var contextParameter = Expression.Parameter(typeof(IStepExecutionContext), "context");
                 var sourceExpr = DynamicExpressionParser.ParseLambda(new [] { dataParameter, contextParameter }, typeof(object), input.Value);
-                var targetExpr = Expression.Property(Expression.Parameter(stepType), input.Key);
 
-                step.Inputs.Add(new DataMapping()
+                var stepParameter = Expression.Parameter(stepType, "step");
+                var targetProperty = Expression.Property(stepParameter, input.Key);
+                var targetExpr = Expression.Lambda(targetProperty, stepParameter);
+
+                step.Inputs.Add(new DataMapping
                 {
                     Source = sourceExpr,
-                    Target = Expression.Lambda(targetExpr)
+                    Target = targetExpr
                 });
             }
         }
@@ -189,12 +192,28 @@ namespace WorkflowCore.Services.DefinitionStorage
             {
                 var stepParameter = Expression.Parameter(stepType, "step");
                 var sourceExpr = DynamicExpressionParser.ParseLambda(new[] { stepParameter }, typeof(object), output.Value);
-                var targetExpr = Expression.Property(Expression.Parameter(dataType), output.Key);
 
-                step.Outputs.Add(new DataMapping()
+                var dataParameter = Expression.Parameter(dataType, "data");
+                Expression targetProperty;
+                
+                // Check if our datatype has a matching property
+                var propertyInfo = dataType.GetProperty(output.Key);
+                if (propertyInfo != null)
+                {
+                    targetProperty = Expression.Property(dataParameter, propertyInfo);
+                }
+                else
+                {
+                    // If we did not find a matching property try to find a Indexer with string parameter
+                    propertyInfo = dataType.GetProperty("Item");
+                    targetProperty = Expression.Property(dataParameter, propertyInfo, Expression.Constant(output.Key));
+                }
+                var targetExpr = Expression.Lambda(targetProperty, dataParameter);
+
+                step.Outputs.Add(new DataMapping
                 {
                     Source = sourceExpr,
-                    Target = Expression.Lambda(targetExpr)
+                    Target = targetExpr
                 });
             }
         }
--- a/src/WorkflowCore/Services/WorkflowExecutor.cs
+++ b/src/WorkflowCore/Services/WorkflowExecutor.cs
@@ -111,7 +111,7 @@ namespace WorkflowCore.Services
 
                         if (result.Proceed)
                         {
-                            ProcessOutputs(workflow, step, body);
+                            ProcessOutputs(workflow, step, body, context);
                         }
 
                         _executionResultProcessor.ProcessExecutionResult(workflow, def, pointer, step, result, wfResult);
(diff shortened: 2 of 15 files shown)


mcrustk2:
Im certainly interested in this. Thanks for the contribution dr-BEat. Hopefully the PR will be resolved.

dr-BEat:
@danielgerlag I don't want to bother you, but what do you think about this PR? Any ideas for improvements?

danielgerlag:
I am so sorry, I have not had a chance to look at this yet. My house flooded and I'm living in a hotel room with 2 kids. Please bear with me.

dr-BEat:
Oh, I am so sorry! No worries, please take all the time you need.

src/WorkflowCore/Services/DefinitionStorage/DefinitionLoader.cs
```diff
@@ -173,12 +173,15 @@ private void AttachInputs(StepSourceV1 source, Type dataType, Type stepType, Wor
...
-                step.Inputs.Add(new DataMapping()
+                var stepParameter = Expression.Parameter(stepType, "step");
+                var targetProperty = Expression.Property(stepParameter, input.Key);
+                var targetExpr = Expression.Lambda(targetProperty, stepParameter);
+
+                step.Inputs.Add(new DataMapping
                 {
                     Source = sourceExpr,
-                    Target = Expression.Lambda(targetExpr)
+                    Target = targetExpr
```
danielgerlag: Could we add a scenario test the exercises this code with an actual JSON defined workflow that does this?
Would you also be willing to write some unit tests around the `DefinitionLoader` class? (I know we don't have any yet, but that would be awesome)

src/WorkflowCore/Services/WorkflowExecutor.cs
```diff
@@ -175,16 +176,23 @@ private void ProcessInputs(WorkflowInstance workflow, WorkflowStep step, IStepBo
...
             foreach (var output in step.Outputs)
             {
-                var member = (output.Target.Body as MemberExpression);
                 var resolvedValue = output.Source.Compile().DynamicInvoke(body);
                 var data = workflow.Data;
-                var property = data.GetType().GetProperty(member.Member.Name);
-                var convertedValue = Convert.ChangeType(resolvedValue, property.PropertyType);
-                property.SetValue(data, convertedValue);
+                var setter = ExpressionHelpers.CreateSetter(output.Target);
+                var convertedValue = Convert.ChangeType(resolvedValue, setter.Parameters.Last().Type);
```
danielgerlag: Don't we need similar changes for `ProcessInputs`?
Could you also add some test cases for the new functionality in the `WorkflowExecutorFixture`?

src/samples/WorkflowCore.Sample03/PassingDataWorkflow2.cs
```diff
@@ -0,0 +1,38 @@
...
+                .StartWith(context =>
+                {
+                    Console.WriteLine("Starting workflow...");
+                    return ExecutionResult.Next();
+                })
+                .Then<AddNumbers>()
+                .Input(step => step.Input1, data => data["Value1"])
+                .Input(step => step.Input2, data => data["Value2"])
+                .Output(data => data["Value3"], step => step.Output)
+                .Then<CustomMessage>()
```
danielgerlag: Indentation looks confusing

test/WorkflowCore.IntegrationTests/Scenarios/DynamicDataIOScenario.cs
```diff
@@ -0,0 +1,69 @@
...
+using System.Collections.Generic;
+using FluentAssertions;
+using WorkflowCore.Interface;
+using WorkflowCore.Models;
+using WorkflowCore.Testing;
+using Xunit;
+
+namespace WorkflowCore.IntegrationTests.Scenarios
+{
+    public class DynamicDataIOScenario : WorkflowTest<DynamicDataIOScenario.DataIOWorkflow, DynamicDataIOScenario.MyDataClass>
```
danielgerlag: Nice 👍

src/samples/WorkflowCore.Sample03/PassingDataWorkflow2.cs
dr-BEat: Sure I will copy the indentation from the original file.

src/WorkflowCore/Services/DefinitionStorage/DefinitionLoader.cs
dr-BEat: Sure! There is already `WorkflowCore.UnitTests.Services.DefinitionStorage.DefinitionLoaderTests` Should I add them there?

src/WorkflowCore/Services/WorkflowExecutor.cs
dr-BEat: No `ProcessInputs` does not need to be changed since it is already treated as a normal lambda. Only the Outputs are "special" in that they are just property names.

I was actually also thinking about changing both Inputs and Outputs to normal lambdas. That would enable getting rid of a lot of reflection and probably big performance improvements (at least for workflows with a lot of property and data movements).
[I started on it here.](https://github.com/dr-BEat/workflow-core/commit/7927c68f45387c1051b7315558f9808ac320bea6) I would love to hear you opinion about it.

dr-BEat: Oh, and I will also add some more tests to `WorkflowExecutorFixture`.

dr-BEat: I added the test.

src/WorkflowCore/Services/DefinitionStorage/DefinitionLoader.cs
dr-BEat: I added a new unittest that loads a workflow that uses a dynamic datatype, I had to change the conditions a little bit to make it works with the indexer returning object instead of just int or just bool.

test/WorkflowCore.IntegrationTests/Scenarios/DynamicDataIOScenario.cs
dr-BEat: thanks (:

danielgerlag:
@dr-BEat would you mind typing up some release notes?

danielgerlag:
@dr-BEat published to nuget as v1.6.7
Please contact me directly, I'd like to discuss something offline.

PR #210, opened 2018-12-10
Add MySQL Provider

erolkskn:
(This pull request created again because of last one was in wrong branch.)
Hi,

First of all thanks for developing such a good library. I created the MySQL Persistence Provider, but first i want to point out some things:

  1. This provider based on Pomelo.EntityFrameworkCore.MySql v2.0.1, only 2.0.1 version can fit in this project.
   2. I didn't add migration feature, that's because of a critical migration bug in 2.0.1 version of Pomelo's MySql Provider. PomeloFoundation/Pomelo.EntityFrameworkCore.MySql#393
   3. Oracle's EF Core MySQL Provider (MySql.Data.EntityFrameworkCore 6.10.7 ) couldn't be option here, it doesn't support Guid Columns feature and there is no way to make it support that feature, it's based on NET Standard v2.x but it can't be run on .NET Framework 4.6.x either.

Q: How can we support MySQL Provider with Migration feature ?
A: It's not easy neither nor hard, if we update the dependency Microsoft.EntityFrameworkCore to 2.1.x from 2.0.1 it will be solved.

Have a good day.

### DIFF
--- a/WorkflowCore.sln
+++ b/WorkflowCore.sln
@@ -113,7 +113,11 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "WorkflowCore.Sample17", "sr
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "WorkflowCore.QueueProviders.SqlServer", "src\providers\WorkflowCore.QueueProviders.SqlServer\WorkflowCore.QueueProviders.SqlServer.csproj", "{7EDD9353-F5C2-414C-AE51-4B0F1C5E105A}"
 EndProject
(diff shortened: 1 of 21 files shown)


erolkskn:
I asked that question because of EF Core version increment will bring these version increments too:

Microsoft.Extensions.DependencyInjection 2.0.0 to 2.1.0
Microsoft.Extensions.DependencyInjection.Abstractions 2.0.0 to 2.1.0
Microsoft.Extension.Logging 2.0.0 to 2.1.0
Microsoft.Extension.Logging.Abstractions 2.0.0 to 2.1.0
and all other DI and Logging related dependencies to 2.1.0

I mean this changes a lot of project dependencies in the solution.

danielgerlag:
Yes, make it another PR...  also, the projects in the library should all only depend on `Microsoft.Extension.*.Abstractions` rather than the actual implementations?

PR #215, opened 2018-12-15
Workflow life cycle events

danielgerlag:
Hi all,

This PR adds functionality to subscribe to workflow life cycle events (WorkflowStarted, WorkflowComplete, WorkflowError, WorkflowSuspended, WorkflowResumed, etc...)
This can be achieved by either grabbing the EventHub from the IoC container and subscribing to events there, or attach an event on the workflow host class.
This implementation only publishes events to the local node... we will still need to implement a distributed version of the EventHub to solve the problem for multi-node clusters.

### DIFF
--- a/src/WorkflowCore/Services/DefaultProviders/SingleNodeEventHub.cs
+++ b/src/WorkflowCore/Services/DefaultProviders/SingleNodeEventHub.cs
@@ -30,7 +30,7 @@ namespace WorkflowCore.Services
                     }
                     catch (Exception ex)
                     {
-                        _logger.LogWarning(ex, $"Error on event subscriber: {ex.Message}");
+                        _logger.LogWarning(default(EventId), ex, $"Error on event subscriber: {ex.Message}");
                     }
                 }
             });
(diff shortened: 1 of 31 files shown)


src/WorkflowCore/Services/DefaultProviders/SingleNodeEventHub.cs
```diff
@@ -0,0 +1,45 @@
...
+            {
+                foreach (var subscriber in _subscribers)
+                {
+                    try
+                    {
+                        subscriber(evt);
+                    }
+                    catch (Exception ex)
+                    {
+                        _logger.LogWarning($"Error on event subscriber: {ex.Message}");
```
Kahbazi:
```suggestion
                        _logger.LogWarning(ex, $"Error on event subscriber: {ex.Message}");
```

can you log exception here so we have the stack trace if anything goes wrong?

src/WorkflowCore/Services/LifeCycleEventPublisher.cs
```diff
@@ -0,0 +1,80 @@
...
+            _cancellationTokenSource.Cancel();
+            DispatchTask.Wait();
+            DispatchTask = null;
+        }
+
+        private async void Execute()
+        {
+            var cancelToken = _cancellationTokenSource.Token;
+            
+            while (!cancelToken.IsCancellationRequested)
```
Kahbazi: I guess we can use `BlockingCollection` like https://github.com/serilog/serilog-sinks-async/blob/dev/src/Serilog.Sinks.Async/Sinks/Async/BackgroundWorkerSink.cs instead of `ConcurrentQueue`, so we don't have a `while (true)`.

PR #220, opened 2018-12-29
EF Core Version Increment and MySQL Migration Support.

erolkskn:
Hi Again,

Update Notes:

1-) As we discussed earlier in #210  , i have incremented the version of EntityFramework Core to 2.1.0 from 2.0.0. That means all other dependent packages are also updated.
2-) All of the sample projects are targeting **.NET Core 2.0 Runtime** now.
3-) Added Migrations to MySQL Persistence Provider.

I ran all tests in both Windows and Mac OS Environment. All unit and integrations tests are passing.


erolkskn:
Thanks 😄 
I have incremented version numbers of changed providers. It seems like AppVeyor doesn't build MySQL package. I think it's configuration related, could you check it ?

Issue #231, opened 2019-01-08
JsonSerializerSettings need to preserve references handling


paksys:
For Workflow using Sqlite and SqlServer, when entities are passed (esp. with circular references) the whole workflow blows out with no warning or errors reported.  When I debugged the code, it seems the Json Serialization was failing for the Entities which I need to pass to the Step Input.  The SerializerSettings are setup as:

JsonSerializerSettings SerializerSettings = new JsonSerializerSettings()
        {
            TypeNameHandling = TypeNameHandling.All
        };

Simply by adding "PreserveReferencesHandling = PreserveReferencesHandling.Objects" fixed the problem. So the settings look like this:

JsonSerializerSettings SerializerSettings = new JsonSerializerSettings()
        {
            TypeNameHandling = TypeNameHandling.All,
            PreserveReferencesHandling = PreserveReferencesHandling.Objects
        };

Probably good idea to add PreserveReferencesHandling settings throughout the solution.  Thoughts?

PR #232, opened 2019-01-08
Fixed issue #231.

paksys:
Minor change to JsonSerializerSettings.


danielgerlag:
Thank you :)... Have you tested how this will affect json documents that were serialized with the old settings?

PR #233, opened 2019-01-09
Netcore2.2

paksys:
Fixed issue #231.
Updated all projects to .Net Standard 2.0, .Net Core 2.2 and latest libraries.


danielgerlag:
I don't want to upgrade the core libraries unless there is a compelling reason to do so.
As a library, we need to maintain maximum compatibility with existing platforms.

PR #245, opened 2019-01-18
Input / Output

danielgerlag:
@dr-BEat @Kahbazi @ssougnez @w-bonelli 

## Action Inputs / Outputs

Added the action Input & Output overloads on the fluent step builder.

```c#
Input(Action<TStepBody, TData> action);
```

This will allow one to manipulate properties on the step before it executes and properties on the data object after it executes, for example

```c#
Input((step, data) => step.Value1 = data.Value1)
```

```c#
.Output((step, data) => data["Value3"] = step.Output)
```

```c#
.Output((step, data) => data.MyCollection.Add(step.Output))
```

## Breaking changes

The existing ability to assign values to entries in dictionaries or dynamic objects on `.Output` was problematic, since it broke the ability to pass collections on the Output mappings.

```c#
.Output(data => data["Value3"], step => step.Output)
```

This feature has been removed, and it is advised to use the action Output API instead, for example

```c#
.Output((step, data) => data["Value3"] = step.Output)
```

This functionality remains intact for JSON defined workflows.


src/WorkflowCore/Interface/IStepBuilder.cs
```diff
@@ -77,6 +77,13 @@ public interface IStepBuilder<TData, TStepBody>
         /// <returns></returns>
         IStepBuilder<TData, TStepBody> Input<TInput>(Expression<Func<TStepBody, TInput>> stepProperty, Expression<Func<TData, IStepExecutionContext, TInput>> value);
 
+        /// <summary>
+        /// Manipulate properties on the step before the it executes
```
germandb:
```suggestion
        /// Manipulate properties on the step before its executed.
```

dr-BEat:
This is quite nice! We can now avoid all the reflection for workflows that are defined in code, maybe it brings even some performance improvements for fast running workflowsteps.

I wanted to do something very similar when I looked at the problem first, but did not want to introduce breaking changes to the Api.

ssougnez:
Another question... How does it play with the json format of the workflow ? Indeed you could potentially use whatever C# code so can you just put this C# code in the JSON or are there some limitations ?

ssougnez:
I really like the work you did in the documentation of WFC. The samples with the README are really handy and provide information in a quick way. However, the JSON part of WFC is pretty undocumented. Indeed, apart from the small part in the main README file, there is nothing. It would be interesting to provide, in each README file where you showed a specific feature of WFC, the JSON representation of the code used in the sample (or at least some kind of pointer to the format to use for the specific feature).

Keep up the good work ;-)

PR #250, opened 2019-01-23
Enhanced to allow scoped dependencies to be injected into steps

MarioAndron:
Step creation was failing with error "WorkflowCore.Services.WorkflowExecutor:Error: Workflow e8c62d2b-249a-4f25-add5-06771c16b33f raised error on step 0 Message: Cannot resolve 'abc' from root provider because it requires scoped service 'def"

Enhanced code to create a scope for the step execution, so that scoped dependencies can be correctly instantiated before injecting.


danielgerlag:
you are awesome! 😄

PR #282, opened 2019-03-13
Update WorkflowHost.cs

perezsebas:
Method added to get the current state of the protected _shutdown property

### DIFF
--- a/src/WorkflowCore/Services/WorkflowHost.cs
+++ b/src/WorkflowCore/Services/WorkflowHost.cs
@@ -153,5 +153,10 @@ namespace WorkflowCore.Services
             if (!_shutdown)
                 Stop();
         }
+        
+        public bool GetShutdownState()
+        {
+            return _shutdown;
+        }
     }
 }


danielgerlag:
What's your use case for this?

perezsebas:
Hello, thank you for your reply. I am trying to run two workflows, one after the other but I am getting an exception ("Operation is not valid due to the current state of the object.") because when runnning the first one I need to execute workflowHost.Start() but when running the second one when the execution reaches this line of code it throwns this exception.

I saw this property called _shutdown that starts false and then turns true when workflowHost.Start() was executed. So the idea is to have a conditional to check if _shutdown equals true then don't run again workflowHost.Start(). The only problem is _shutdown property is protected so I can't access it, that's why I added this method to be able to get its value.

germandb:
The WorkflowHost.Start() only needs to be invoke once when you start your service. You dont need to start the Host for each Workflow. Please take a look into [Using WorkflowCore with ASP.NET.Core](https://github.com/danielgerlag/workflow-core/wiki/97.-Using-with-ASP.NET-Core)

danielgerlag:
yup, the intended usage is to start the host when you app starts and leave it started as it runs internal worker pools.

perezsebas:
Thank you guys for your help, it is working like you said, have a nice week.

PR #285, opened 2019-03-20
WorkflowOptions support for configuring number of active tasks by multipler

eddie3716:
The idea is the api consumer can control the number of active tasks through Workflow Options configuration.  They pass in a multiplier value, constrained to be between 1 through 10, which is then used by the QueueConsumer to compute the MaxConcurrentItems field.  This is useful in situations where changing the processor core count in order to generate more active current items is not practical or possible.


eddie3716:
@danielgerlag thanks for the feedback.  I started over from scratch (deleted my repo, reforked, etc)  and created pull request 287.  Can you close out this merge request and use the other one instead?

eddie3716:
Close.  Would rather use pull request 287

PR #287, opened 2019-03-25
WorkflowOptions support for configuring max number of concurrent tasks.

eddie3716:


SebastienLagrange:
Hello,
I am very happy to see that you implemented this change that I was just thinking on !
I implement workflows that contains lot of steps that actually are waiting for external resources. And I start hundreds of workflows at the same time. The current implementation limiting the number of tasks to the number of CPU is very limiting the performance to process all the hundreds of workflows... and I see my CPU is zero all the time because the steps actually call REST services that are on other server so the workflow engine is just waiting all the time. In that case it would possible to start 50 workflows at the same time even with my 2 CPU, and that would really improve the time to process all the hundreds of workflows...
daniel do you think this modification is good, will you accept the pullrequest and make this new option available from your master package ?
Thank you all !

src/WorkflowCore/Models/WorkflowOptions.cs
```diff
@@ -67,6 +70,16 @@ public void UseErrorRetryInterval(TimeSpan interval)
         {
             ErrorRetryInterval = interval;
         }
+
+        public void UseMaxConcurrentItems(int? maxConcurrentItems)
```
danielgerlag: Perhaps we should make this explicitly the max number of concurrent workflows rather than items?

src/WorkflowCore/Services/BackgroundTasks/EventConsumer.cs
```diff
@@ -13,7 +13,7 @@ internal class EventConsumer : QueueConsumer, IBackgroundTask
         private readonly IPersistenceProvider _persistenceStore;
         private readonly IDistributedLockProvider _lockProvider;
         private readonly IDateTimeProvider _datetimeProvider;
-        protected override int MaxConcurrentItems => 2;
+        protected override int MaxConcurrentItems => WorkflowOptions.MinimumNumberOfConcurrentItems;
```
danielgerlag: This doesn't make sense to me... is this a typo?

eddie3716: That was intentional, because MaxConcurrentItems  as a function of the number 2 was showing in two places, the EventConsumer and QueueConsumer.  And then I added a 3rd place it was showing up, with the WorkflowOptions.  I was concerned the number 2 was becoming a magic number, but this change doesn't add a ton of value, so I can revert it if its problematic.

src/WorkflowCore/Models/WorkflowOptions.cs
eddie3716: Let me make that change you've suggested.  Most consumers of your API may not ever dig into the internals, and thus not know what maxConcurrentItems really means, but framing it as "max number of concurrent workflows" is more intuitive.

eddie3716: I think instead of modifying EventConsumer directly, I'll move the change to WorkflowConsumer...that way the other child consumers can keep their default behavior the way you intended.

PR #290, opened 2019-03-26
chore(sample): Improve WebApiSample, adding SwaggerUI, fix invalid cast from JObject

hieuxlu:
EventsController in WebApiSample is receiving event data as JObject, but TestWorkFlow is trying to cast data to string. This would probably cause headache when first opening the WebApiSample.


hieuxlu:
Cast JObject on publish event will not work with different workflow data type.

danielgerlag:
As this is a sample, I'd prefer to keep it simple, it does not need to work for all sorts of types... just illustrate how it works.

hieuxlu:
Sorry for late reply. Yes, for simplicity, I've moved the cast to EventController. I've also added Swagger for easy debugging WebApiSample. 

Please have a look. Thank you.

PR #323, opened 2019-06-13
EF WorkflowInstance.ToPersistable change to not recreate ExecutionPointers

titns:
A fix for an issue where on each workflow persist all executionpointers were re-created.
Creation of new persistance.ExecutionPointers collection object causes the EntityFramework to delete existing previous records in database on SaveChanges.
The fix causes existing records to be untouched if no changes occur, otherwise the existing records are updated (instead of insert).

### DIFF
--- a/src/providers/WorkflowCore.Persistence.EntityFramework/ExtensionMethods.cs
+++ b/src/providers/WorkflowCore.Persistence.EntityFramework/ExtensionMethods.cs
@@ -27,7 +27,7 @@ namespace WorkflowCore.Persistence.EntityFramework
             persistable.Status = instance.Status;
             persistable.CreateTime = instance.CreateTime;
             persistable.CompleteTime = instance.CompleteTime;
-            persistable.ExecutionPointers = persistance.ExecutionPointers ?? new PersistedExecutionPointerCollection(instance.ExecutionPointers.Count);
+            persistable.ExecutionPointers = persistable.ExecutionPointers ?? new PersistedExecutionPointerCollection(instance.ExecutionPointers.Count);
             
             foreach (var ep in instance.ExecutionPointers)
             {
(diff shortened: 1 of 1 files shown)


src/providers/WorkflowCore.Persistence.EntityFramework/ExtensionMethods.cs
```diff
@@ -27,7 +27,7 @@ internal static PersistedWorkflow ToPersistable(this WorkflowInstance instance,
             persistable.Status = instance.Status;
             persistable.CreateTime = instance.CreateTime;
             persistable.CompleteTime = instance.CompleteTime;
-            persistable.ExecutionPointers = new PersistedExecutionPointerCollection(instance.ExecutionPointers.Count);
+            persistable.ExecutionPointers = persistable.ExecutionPointers ?? new PersistedExecutionPointerCollection(instance.ExecutionPointers.Count);
```
danielgerlag: How will this behave when new execution pointers are added?
Did you run this against the SQL integration tests?

titns: When new execution pointers are added, the method will work as usual, i.e. peristable.ExecutionPointers.FindById(ep.id) at line 32 will return null, therefore a new pointer will be added to the collection. That pointer will be caught by (EF query).AsTrackable() and will be translated into an insert into the database.

From internal tests in a personal project, all workflow instances run/complete as expected on a workflow with multiple parallel branches containing while loops and expecting events. Only difference is a heavily reduced performance impact.

Integration tests/Sql Server unit tests are passing:
image

danielgerlag: I need to confirm this behavior

PR #333, opened 2019-07-02
Fixing issue with waiting for same event a 2nd time on the same step.

gthomps95:
I want to allow an event to fire a workflow in the same step more than once, so at the end of a step, I wait for the same event again.  This was not working after the first event.  I did some debugging and to me it looks like this change is required for this to work.  Don't know enough about the usage of the EventPublished property to know if there are ramifications of doing this.  But, I think if you're potentially changing the name and key on the pointer, it's also safe to change the EventPublished flag.

I added an integration test scenario.  I ran all the integration and unit tests and they pass.

### DIFF
--- a/src/WorkflowCore/Services/ExecutionResultProcessor.cs
+++ b/src/WorkflowCore/Services/ExecutionResultProcessor.cs
@@ -41,7 +41,6 @@ namespace WorkflowCore.Services
             {
                 pointer.EventName = result.EventName;
                 pointer.EventKey = result.EventKey;
-                pointer.EventPublished = false;
                 pointer.Active = false;
                 pointer.Status = PointerStatus.WaitingForEvent;
 
@@ -55,6 +54,11 @@ namespace WorkflowCore.Services
                 });
             }
 
+            if (result.EventReWait)
+            {
+                pointer.EventPublished = false;
+            }
+
             if (result.Proceed)
             {
                 pointer.Active = false;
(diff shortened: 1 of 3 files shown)


src/WorkflowCore/Services/ExecutionResultProcessor.cs
```diff
@@ -41,6 +41,7 @@ public void ProcessExecutionResult(WorkflowInstance workflow, WorkflowDefinition
             {
                 pointer.EventName = result.EventName;
                 pointer.EventKey = result.EventKey;
+                pointer.EventPublished = false;
```
danielgerlag: I believe this would cause existing workflows to behave unexpectedly - we should probably write some tests to guard against that... under normal circumstances, a `WaitFor` should only fire once.

You could implement your own alternate `ReWaitFor` primitive... have a look at https://github.com/danielgerlag/workflow-core/blob/master/src/WorkflowCore/Primitives/WaitFor.cs

When it returns `ExecutionResult.Next()`, it will no longer wait for any further events, but when it returns `ExecutionResult.WaitForEvent(EventName, EventKey, effectiveDate)`, it will wait for more... you can adapt this to your needs.  You can also look at the fluent builder extension method to see how to call it - https://github.com/danielgerlag/workflow-core/blob/master/src/WorkflowCore/Services/FluentBuilders/StepBuilder.cs#L133

Let me know if you need more info.

gthomps95: I don't think only doing the primitive will work due to line 89 in EventConsumer, if the event published flag doesn't get reset, the subscription doesn't get built for the next event.  Also, I don't think this method will make sense in the fluent builder since there are whiles to make this scenario work.  I think it's only a response to a custom step, that I need to wait for the event again.  Does that make sense?

I will commit a different change that will isolate the logic to only be for a call to ExecutionResult.ReWaitFor(...).

Let me know your thoughts.

PR #340, opened 2019-07-10
Adding null check for missing input step property name

gthomps95:
Adding null check for missing input step property name when loading definition from storage to give better error message.


test/WorkflowCore.TestAssets/Utils.cs
```diff
@@ -32,6 +32,11 @@ public static string GetTestDefinitionDynamicJson()
         {
             return File.ReadAllText("stored-dynamic-definition.json");
         }
+
+        public static string GetTestDefinitionJsonInputExc()
+        {
+            return File.ReadAllText("stored-definition-input-exc.json");
```
danielgerlag: What does `-exc` stand for?  Could we make it more explicit?

gthomps95: Ok.  I changed the file name and appropriate methods/files.

PR #354, opened 2019-07-24
Fix issue # 332

courosh12:
Made the schema name parameterized without changing too much. I can also make method overloads if this solution is less optimal


danielgerlag:
Thank you :)
Can you increment the version numbers of all the libraries that need to be deployed?

courosh12:
yes ofcource i think its only the Postgres client

courosh12:
is this correct? not sure how the ci/cd is set up

Issue #358, opened 2019-07-29
Workflow executes interrupt sometimes


qazq:
Hi,

After I upgrade workflow-core to the latest version, my workflow sometimes executes interrupt until the RunnablePoller put the task into the queue again. The worst-case should wait for 10 seconds to execute the next step. I try to find out the issue, it seems caused by the secondPasses feature in the QueueConsumer.

Here is my test code: https://github.com/qazq/workflow-core/commit/b9ada304abafaff3a0418f922fc143a44ea3c3db

I use the Sample10 WhileWorkflow. The output:
```txt
Starting workflow...
info: WorkflowCore.Services.WorkflowHost[0]
      Starting background tasks
info: WorkflowCore.Services.BackgroundTasks.RunnablePoller[0]
      Polling for runnable workflows
info: WorkflowCore.Services.BackgroundTasks.RunnablePoller[0]
      Polling for unprocessed events
Hello
Doing something...
Doing something...
Doing something...
Doing something...
Doing something...
Doing something...
Doing something...
Doing something...
Doing something...
Doing something...
Doing something...
Doing something...
Doing something...
Doing something...
Doing something...
Doing something...
Doing something...
Doing something...
Doing something...
Doing something...
Doing something...
Doing something...
Doing something...
Doing something...
Doing something...
Doing something...
Doing something...
Doing something...
Doing something...
Doing something...
Doing something...
Doing something...
Doing something...
Doing something...
Doing something...
Doing something...
Doing something...
Doing something...
Doing something...
Doing something...
Doing something...
Doing something...
Doing something...
Doing something...
Doing something...
Doing something...
Doing something...
Doing something...
Doing something...
Doing something...
Doing something...
Doing something...
Doing something...
Doing something...
>>>>>>>>> secondPasses Workflow, activeTasks.ContainsKey(item): True

// here wait for 8 seconds....

info: WorkflowCore.Services.BackgroundTasks.RunnablePoller[0]
      Polling for runnable workflows
info: WorkflowCore.Services.BackgroundTasks.RunnablePoller[0]
      Got runnable instance 53a3daba-fd99-43b8-93c2-8f12b0e82802
info: WorkflowCore.Services.BackgroundTasks.RunnablePoller[0]
      Polling for unprocessed events
Doing something...
Doing something...
Doing something...
Doing something...
Doing something...
Doing something...
>>>>>>>>> secondPasses Workflow, activeTasks.ContainsKey(item): True

// here wait for 10 seconds....

[...]

PR #366, opened 2019-08-12
Adjust the timing of put the task into the queue

qazq:
make sure the active tasks in the queue consumer has removed before putting the future task into the queue.

Fix bug #358


danielgerlag:
How about this?
https://github.com/danielgerlag/workflow-core/commit/0931aab4ba6756fe6b883f9511175a476e41b9b0

qazq:
It works fine, thanks a lot.

PR #371, opened 2019-08-18
Added StepExecutionContextAccessor

emyklebost:
**Purpose**
My team needed the ability to get the StepExecutionContext through DI. Specifically to be used within a CorrelationIdDelegationHandler class, which populates the X-Correlation-Id header with the executing Workflow-Instance-Id.

**Implementation**
The implementation uses a StepExecutionContextAccessor based on the HttpContextAccessor from ASP.NET Core, which uses AsyncLocal to store the context. This is needed in order to be used within Http DelegationHandlers, because when using HttpClientFactory the HttpClient and DelegationHandlers are cached and thereby not tied to the DI lifetime.

### DIFF
--- a/src/WorkflowCore/Services/WorkflowExecutor.cs
+++ b/src/WorkflowCore/Services/WorkflowExecutor.cs
@@ -141,6 +141,21 @@ namespace WorkflowCore.Services
             {
                 _logger.LogDebug("Starting step {0} on workflow {1}", step.Name, workflow.Id);
 
+                IStepExecutionContext context = new StepExecutionContext()
+                {
+                    Workflow = workflow,
+                    Step = step,
+                    PersistenceData = pointer.PersistenceData,
+                    ExecutionPointer = pointer,
+                    Item = pointer.ContextItem
+                };
+
+                if (scope.ServiceProvider.GetService<IStepExecutionContextAccessor>()
+                    is StepExecutionContextAccessor accessor)
+                {
+                    accessor.StepExecutionContext = context;
+                }
+
                 IStepBody body = step.ConstructBody(scope.ServiceProvider);
 
                 if (body == null)
@@ -156,15 +171,6 @@ namespace WorkflowCore.Services
                     });
                     return;
                 }
-
-                IStepExecutionContext context = new StepExecutionContext()
-                {
-                    Workflow = workflow,
-                    Step = step,
-                    PersistenceData = pointer.PersistenceData,
-                    ExecutionPointer = pointer,
-                    Item = pointer.ContextItem
-                };
 
                 foreach (var input in step.Inputs)
                     input.AssignInput(workflow.Data, body, context);
--- a/src/WorkflowCore/Interface/IStepExecutionContextAccessor.cs
+++ b/src/WorkflowCore/Interface/IStepExecutionContextAccessor.cs
@@ -1,0 +1,7 @@
+﻿namespace WorkflowCore.Interface
+{
+    public interface IStepExecutionContextAccessor
+    {
+        IStepExecutionContext StepExecutionContext { get; }
+    }
+}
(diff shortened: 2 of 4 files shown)


emyklebost:
> How come you couldn't just pass the data you need in as a parameter from the calling step to your dependency? That way it also wouldn't couple your dependencies to workflow core?

In this case the CorrelationIdDelegationHandler does not directly depend on Workflow-core, but rather a IUserSession abstraction that has multiple implementations; E.g. HttpSession:IUserSession, which uses the HttpContext and a WorkflowSession:IUserSession, which uses the StepExecutionContext.

danielgerlag:
What part of the execution context do you need to access?

emyklebost:
> What part of the execution context do you need to access?

In the case I've descriped we access metadata about the Workflow and Step (WorkflowInstance.Id & WorkflowStep.Name) for cross-cutting concerns like populating log messages and headers in HttpRequests so that we are able to associate logs, metrics and HttpRequests with a particular workflow and\or step.

However, in general, such a feature would provide access to the StepExecutionContext when using DI to construct Services, including factory-methods that returns scoped/transient service-instances depending on the current context (e.g. IUserSession example in my previous comment).

_Note_: I did actually start off adding the IStepExecutionContextAccessor as a Soped service within the StepExecution DI scope, but changed it to follow the HttpContextAccessor pattern (singleton with AsyncLocal) to make it usable with HttpClientFactory.

danielgerlag:
It makes sense for cross cutting concerns, like logging and instrumentation.  For actual business logic I would want to discourage people from passing context down to their dependencies likes this... how could we achieve that as well?

emyklebost:
> It makes sense for cross cutting concerns, like logging and instrumentation. For actual business logic I would want to discourage people from passing context down to their dependencies likes this... how could we achieve that as well?

I agree that such a feature can be used in ways that might not be considered best practice, but shouldn't how it is used ultimately be up to the user? E.g. the HttpContext in an ASP.NET application can be accessed by any service participating in handling a request, that power can definitely be misused and\or lead to bad software design.

If the user chooses to follow best practices of separating business and infrastructure logic he would probably have an architecture that prevents business logic from accessing the StepExecutionContext or HttpContext directly, and instead depend on abstractions (ref. DDD\Clean\Ports-and-adapters\Onion -architecture etc.).

danielgerlag:
While that is true, I like to make an effort to think about how the design of a library can guide even inexperienced developers down a good path.

src/WorkflowCore/Services/WorkflowExecutor.cs
```diff
@@ -141,6 +141,21 @@ private async Task ExecuteStep(WorkflowInstance workflow, WorkflowStep step, Exe
...
+                    Step = step,
+                    PersistenceData = pointer.PersistenceData,
+                    ExecutionPointer = pointer,
+                    Item = pointer.ContextItem
+                };
+
+                if (scope.ServiceProvider.GetService<IStepExecutionContextAccessor>()
+                    is StepExecutionContextAccessor accessor)
+                {
+                    accessor.StepExecutionContext = context;
```
danielgerlag: I think, instead of baking this into the core library... I would put an event that the user can hook into here... something like `OnExecutionScopeConstruction` or something like that.  Then the user can customize the execution scope to have whatever they want.

emyklebost: In that case, why not introduce some sort of StepExecutionFilter; a way of adding code that will run before and\or after every Step within one or multiple Workflows? Similar to Filter\Middleware in ASP.NET or Pipeline-behavior i Mediatr. This is far beyond what this branch was trying to solve, but would be a very powerful feature for implementing cross-cutting concerns.

danielgerlag: Sorry, I missed this update... that sounds like an interesting idea!

maulik-modi: Yes,  I agree Middleware would be a great feature to have

shirados90:
@emyklebost  Would these changes help you to achieve your goal? #627

PR #396, opened 2019-09-10
Bugfix for accessing nested properties using the Json/Yaml notation.

OAlbrecht2407:
When using the c#-fluent-interface you can access subproperties of the data-object  within the Output(...) extension (e.g.: Output(s =>s.SubData.ExecuteIf, s => s.IsDigit). Using the Json/Yaml notification, this isn't possible because the DefinitionLoader assumes that a property of the data-object is accessed.

With my changes the DefinitionLoader checks if there a points ('.') in the expression and if so, it try to map it to a subproperty.

### DIFF
--- a/src/WorkflowCore/Services/DefinitionStorage/DefinitionLoader.cs
+++ b/src/WorkflowCore/Services/DefinitionStorage/DefinitionLoader.cs
@@ -208,6 +208,33 @@ namespace WorkflowCore.Services.DefinitionStorage
 
                 var dataParameter = Expression.Parameter(dataType, "data");
                 Expression targetProperty;
+                
+                // If the property has a point, check for a nested type
+                Expression parentExpression = dataParameter;
+                Type parentDataType = dataType;
+                string key = output.Key;
+                bool continueLoop = key.Contains(".");
+                while (continueLoop)
+                {
+                    var indexOfPoint = key.IndexOf(".", StringComparison.InvariantCultureIgnoreCase);
+                    var subPropertyName = key.Substring(0,indexOfPoint);
+                    var subProperty = parentDataType.GetProperty(subPropertyName);
+                    key = key.Substring(indexOfPoint + 1);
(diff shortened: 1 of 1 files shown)


sorengranfeldt:
+1 for getting this merged

coderookie1994:
Is anyone looking into this ? This will be really good if it's accepted.

src/WorkflowCore/Services/DefinitionStorage/DefinitionLoader.cs
```diff
@@ -208,6 +208,33 @@ private void AttachOutputs(StepSourceV1 source, Type dataType, Type stepType, Wo
 
                 var dataParameter = Expression.Parameter(dataType, "data");
                 Expression targetProperty;
+                
+                // If the property has a point, check for a nested type
+                Expression parentExpression = dataParameter;
```
danielgerlag, line 213: Could we encapsulate this block in a private method?

src/WorkflowCore/Services/DefinitionStorage/DefinitionLoader.cs
```diff
@@ -208,6 +208,33 @@ private void AttachOutputs(StepSourceV1 source, Type dataType, Type stepType, Wo
...
+                    var subPropertyName = key.Substring(0,indexOfPoint);
+                    var subProperty = parentDataType.GetProperty(subPropertyName);
+                    key = key.Substring(indexOfPoint + 1);
+                    continueLoop = key.Contains(".");
+                    var subPropertyParameterProperty = Expression.Property(parentExpression, subProperty);
+                    parentExpression = subPropertyParameterProperty;
+                    parentDataType = subProperty.PropertyType;
+                    if(!continueLoop)
+                    {
+                        subPropertyName = key;
```
danielgerlag, line 229: Is this a duplication of the original logic below?

danielgerlag (changes requested):

PR #404, opened 2019-09-25
Add Azure ServiceBus provider

glucaci:
Added event hub support using Azure ServiceBus provider.

### DIFF
--- a/src/providers/WorkflowCore.Providers.Azure/README.md
+++ b/src/providers/WorkflowCore.Providers.Azure/README.md
@@ -22,13 +22,13 @@ dotnet add package WorkflowCore.Providers.Azure
 ## Usage
 
 Use the `IServiceCollection` extension methods when building your service provider
-* .UseAzureSyncronization
-* .UseAzureServiceBus
+* .UseAzureSynchronization
+* .UseAzureServiceBusEventHub
 
 ```C#
 services.AddWorkflow(options => 
 {
-	options.UseAzureSyncronization("azure storage connection string");
-	options.UseAzureServiceBus("service bus connection string", "topic name", "subscription name");
+	options.UseAzureSynchronization("azure storage connection string");
+	options.UseAzureServiceBusEventHub("service bus connection string", "topic name", "subscription name");
 });
 ```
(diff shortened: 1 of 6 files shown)


glucaci:
I don't think so, Azure Service Bus is a publish/subscriber service like Redis and on the other side Azure Event Hub is a event streaming service like Kafka.

danielgerlag:
Exactly, the first event notification provider for WorkflowCore was built on AWS Kinesis - https://github.com/danielgerlag/workflow-core/tree/master/src/providers/WorkflowCore.Providers.AWS
Service bus would be more suited to the queue provider rather than event notification.

glucaci:
Nice, than could we merge this ?
Thanks !

src/providers/WorkflowCore.Providers.Azure/README.md
```diff
@@ -2,21 +2,33 @@
...
+Use the `IServiceCollection` extension methods when building your service provider
+* .UseAzureSyncronization
+* .UseAzureServiceBus
 
 ```C#
-services.AddWorkflow(x => x.UseAzureSyncronization("azure storage connection string"));
+services.AddWorkflow(options => 
+{
+	options.UseAzureSyncronization("azure storage connection string");
+	options.UseAzureServiceBus("service bus connection string", "topic name", "subscription name");
```
danielgerlag: I not sure about the name of this extension method.  I think we we be explicit of what Azure Service bus is being used for... there is ambiguity between a queue provider and event hubs here.  Then it also leaves to door open to doing an implementation with Azure Event Hubs, which would be a more natural choice for this use case, I think.

glucaci: If we introduce the Workflow-Core abstraction naming in extension methods will look not so nice. E.g. "UseAzureServiceBusEventHub" and for Azure EventHub implementation will look something like "UseAzureEventHubEventHub". I think through documentation user can find out what involves by using this.
Also there is some inconsistency because Redis is using it in the extension method naming and Aws Kinesis not.
Maybe will be nice not to propagate internal abstraction naming to the client and just document it.

danielgerlag: I do not agree that the explicit naming propagates an internal abstraction.  I think the consumer of the library should be able to control exactly what external provider is used for what.  They should be able to pick and choose between each provider for specific uses.

glucaci: Done

glucaci:
@danielgerlag could you please recheck. Thanks :)

PR #443, opened 2019-11-06
Work Flow Ended Events

NathanImpreza:
This adds events when a workflow ends.

These can be used for scenarios where one workflow starts another and the initial work flow needs to wait for the sub workflow to complete.

This is my first pull request please let me know if there is anything I need to do or change.


NathanImpreza:
Okay after reading through the release information I've subscribed to the lifeCycle events and can publish my workflow end event from there instead. 

Thank you for that I'll close this pull request.

PR #444, opened 2019-11-06
Ability to add data directly to a workflow

NathanImpreza:
This method can be used to directly update workflow data, we use this function via a REST api to add state data to the work flow instance allowing us to use that data in later steps and for reporting purposes. 

This can be useful when calling out to an extenal process (e.g. a REST API) from the workflow, where the external process may do multiple functions and need to push data back to the workflow during any one of those steps. Or when there is a long running function external to the workflow, but the work flow should be updated to that state of that function.

E.g. We have a have a workflow which kicks off a survey.
The survey has multiple pages and can be saved and returned to over a number of weeks.
We use this method to store the current state and page number of the survey so that returning user is displayed the correct page, and so that the administrators can see where in the process the user is.

We can also push some survey data into the workflow and use IF steps to change outcomes based on the survey data.

We could do this using events and multiple steps one waitForEvent per page, but that would tightly couple our dynamic survey system to the workflow steps. Therefore the approch of allowing direct data access seems to make sense.

Hopefuly this scenario makes sense nad you can see the use case forthis functionality.

Please let me know if this pull request needs any more information or any changes.

### DIFF
--- a/src/WorkflowCore/Interface/IWorkflowController.cs
+++ b/src/WorkflowCore/Interface/IWorkflowController.cs
@@ -12,7 +12,8 @@ namespace WorkflowCore.Interface
         Task<string> StartWorkflow<TData>(string workflowId, TData data = null, string reference=null) where TData : class, new();
         Task<string> StartWorkflow<TData>(string workflowId, int? version, TData data = null, string reference=null) where TData : class, new();
 
-        Task PublishEvent(string eventName, string eventKey, object eventData, DateTime? effectiveDate = null);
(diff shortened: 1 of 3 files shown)


NathanImpreza:
After reading your comment and doign a bit of research I can see what you mean, I think we were trying to store too much in the workflow and so it was becoming more of a data store and less of a state store we've refactored the survey code and it can now work without this data being directly updated into the workflow. 

Thankyou for the information, I'll close this pull request.

PR #445, opened 2019-11-09
upper version constraints against ef core 3

danielgerlag:
* EF Core 3.0 introduces [breaking changes](https://github.com/aspnet/EntityFramework.Docs/blob/master/entity-framework/core/what-is-new/ef-core-3.0/breaking-changes.md#provider)
* EF Core 3.0 targets .net standard 2.1 which will [not be supported by .Net Framework](https://devblogs.microsoft.com/dotnet/announcing-net-standard-2-1/)
* This PR places an upper limit on the EF core version of the host application, as this incompatibility was creating confusion that workflow core is not compatible with .net core 3.0, which is false


terry-delph:
HI @danielgerlag 

I think the MySQL persistence provider may also need this same constraint.

It depends on **_Pomelo.EntityFrameworkCore.MySql_**.

If we were to upgrade to [Pomelo.EntityFrameworkCore.MySql 3.0.0-rc3.final](https://www.nuget.org/packages/Pomelo.EntityFrameworkCore.MySql/3.0.0-rc3.final)

Dependency chain of Pomelo is:

-  _**Microsoft.EntityFrameworkCore.Relational**_ >= 3.0.0: [Microsoft.EntityFrameworkCore.Relational/](https://www.nuget.org/packages/Microsoft.EntityFrameworkCore.Relational/)

   - **_Microsoft.EntityFrameworkCore_** >= 3.0.0

PR #461, opened 2019-12-03
Replce mongo obsolate api and add enum to string convetion

glucaci:
- Replace create index obsoleted api
- Add convention pack to serialize Enum's as string value (more readable)

### DIFF
--- a/src/providers/WorkflowCore.Persistence.MongoDB/Services/MongoPersistenceProvider.cs
+++ b/src/providers/WorkflowCore.Persistence.MongoDB/Services/MongoPersistenceProvider.cs
@@ -72,8 +72,14 @@ namespace WorkflowCore.Persistence.MongoDB.Services
         {
             if (!indexesCreated)
             {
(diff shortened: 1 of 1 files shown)


src/providers/WorkflowCore.Persistence.MongoDB/Services/MongoPersistenceProvider.cs
```diff
@@ -24,6 +26,13 @@ public MongoPersistenceProvider(IMongoDatabase database)
 
         static MongoPersistenceProvider()
         {
+            ConventionRegistry.Register(
+                "workflow.conventions",
+                new ConventionPack
+                {
+                    new EnumRepresentationConvention(BsonType.String)
+                }, t => t.FullName?.StartsWith("WorkflowCore") ?? false);
```
danielgerlag: is this compatible with existing databases out there that have done it the old way?

glucaci: Yes, it can serialize/deserialize in both ways. The new or replaced documents will have the new representation and the old ones will remain as they are.

PR #470, opened 2019-12-10
Enable override of RunAsync in StepBody

mfeerick:
Allow implementations of StepBody to override the RunAsync method so asynchronous operations can be performed with the await keyword inside a StepBody.

Removed the abstract keyword from the Run method so implementors overriding RunAsync do not need to override both Run and RunAsync.
Implementors who choose not to override RunAsync do not need to make any changes as the default implementation of RunAsync makes a call to Run.
Implementors who forget to override either Run or RunAsync will receive a NotImplementedException instead of the compiler abstract method warning.

### DIFF
--- a/src/WorkflowCore/Models/StepBody.cs
+++ b/src/WorkflowCore/Models/StepBody.cs
@@ -6,9 +6,12 @@ namespace WorkflowCore.Models
 {
     public abstract class StepBody : IStepBody
     {
-        public abstract ExecutionResult Run(IStepExecutionContext context);
+        public virtual ExecutionResult Run(IStepExecutionContext context)
+        {
+            throw new NotImplementedException();
+        }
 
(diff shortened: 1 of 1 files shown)


ssougnez:
Maybe I misunderstood the goal of this but why don't you simply inherits from StepBodyAsync ?

mfeerick:
Yes you are right.

I didn’t see that there was a separate StepBodyAsync.

Curiously there are more methods in the non async version.


On 10 Dec 2019, at 17:41, Sébastien Sougnez <notifications@github.com> wrote:

﻿

Maybe I misunderstood the goal of this but why don't you simply inherits from StepBodyAsync ?

mfeerick:
Yes you are right, I didnt see there was a separate StepBodyAsync.

Curiously there is more code in the non async version.

PR #485, opened 2020-01-03
Updated the Pomelo library to 3.1.0 to make it compatible with Entity…

paulose:
…Framework.Core 3.1.0 and bumped the WorkflowCore.Persistence.MySQL version to 1.2.2


paulose:
No. It does not need a new migration. I have tested this on my local with the existing migration and submitted a new workflow and it worked.

paulose:
Is there any other check you would like me to perform?

danielgerlag:
The MySQL provider doesn't have the latest 2 migrations that all the other EF providers have.
Some of the new features may not work,

danielgerlag:
updated in another PR

Issue #431, opened 2019-10-25
If in ForEach loop


pscheicher:
Hi,
is it possible to have an If statement inside a ForEach loop, where the condition is based on the context.Item.
I hope this example clarifies the problem:
```
builder.StartWith<GenerateObjectsStep>()
    .Output(data => data.GeneratedObjectsList, step => step.GeneratedObjectsList)
    .ForEach(data => data.GeneratedObjectsList)
       .Do(branch =>
           branch.StartWith<AnalyzeObject>()
                     .Input(step => step.GeneratedObject, (data, context) => context.Item)
                     .If(data => data.IsObjectPrintable).Do(then => then
                          .StartWith<PrintObject>()
                          .Input(step => step.GeneratedObject, (data, context) => context.Item)
                     )
            )
            .Then<SaveObject>();
       )
```
I would like to execute several steps (in this example only one) based on the objects generated during the workflow.

danielgerlag:
The JSON / YAML definitions do currently support this, but the fluent API does not have it exposed.... this is something we could add.

pscheicher:
Thank you for your feedback.
Addning this feature would be great. For now I check at the begin of each of those steps, if the condition is matched and abort the execution. I will have a look on the JSON definitions. But I would prefer using it in fluent API.

ChrisBertrand:
Has this been added?

IvanJosipovic:
@ChrisBertrand, no, I have a PR for it https://github.com/danielgerlag/workflow-core/pull/1095

PR #505, opened 2020-02-04
Implement #431

sebastianmattar:
I extended to Fluent API to allow access to the IStepExecutionContext in If() and Output().
This allows us to access the current item in ForEach() blocks.

Can you have a look? Thank you!


sebastianmattar:
Sure, I added a unit test to cover the new functionality.

src/WorkflowCore/Models/MemberMapParameter.cs
```diff
@@ -44,8 +43,19 @@ private void Assign(object sourceObject, LambdaExpression sourceExpr, object tar
             }
 
             var valueExpr = Expression.Convert(Expression.Constant(resolvedValue), targetExpr.ReturnType);
-            var assign = Expression.Lambda(Expression.Assign(targetExpr.Body, valueExpr), targetExpr.Parameters.Single());
-            assign.Compile().DynamicInvoke(targetObject);
+            var assign = Expression.Lambda(Expression.Assign(targetExpr.Body, valueExpr), targetExpr.Parameters);
+
+            switch (targetExpr.Parameters.Count)
```
danielgerlag: Not sure I understand why we're doing this?

test/WorkflowCore.IntegrationTests/Scenarios/StepContextScenario.cs
```diff
@@ -0,0 +1,109 @@
...
+                            x => x.StartWith<AddNumbers>()
+                                .Input(step => step.Input1, data => data.Sum)
+                                .Input(step => step.Input2, (data, ctx) => ((LoopIteration)ctx.Item).ToAdd)
+                                .Output((data, ctx) => ((LoopIteration)ctx.Item).Result, step => step.Output)
+                                .Output(data => data.Sum, step => step.Output)
+                            .If((step, ctx) => ((LoopIteration)ctx.Item).ToAdd > 3)
+                                .Do(y => y.StartWith<AddNumbers>()
+                                    .Input(step => step.Input1, data => data.GreaterThree)
+                                    .Input(step => step.Input2, data => 1)
+                                    .Output(data => data.GreaterThree, step => step.Output)
```
danielgerlag: The `Output` mapping doesn't seem intuitive?

src/WorkflowCore/Services/FluentBuilders/StepBuilder.cs
```diff
@@ -335,7 +341,22 @@ private WorkflowStep IterateParents(int id, string name)
 
             return stepBuilder;
         }
-        
+
+        public IContainerStepBuilder<TData, If, If> If(Expression<Func<TData, IStepExecutionContext, bool>> condition)
+        {
+            var newStep = new WorkflowStep<If>();
+
+            Expression<Func<If, bool>> inputExpr = (x => x.Condition);
```
danielgerlag: Not sure this is going to work... we're losing the context parameter here

sebastianmattar:
Alright, let me explain first what I want to achieve:
I want to implement a "parallel approval" workflow that waits for user feedback and updates the workflow state accordingly. To make this work I need to have a way to use the item context as output parameter.

I implemented it like this:
`builder
	.StartWith<SayHello>()
	.ForEach(data => data.ParallelApprovals)
		.Do(x => x
			.StartWith<GetUserFeedback>()
				.Output((data, context) => ((MyWorkflowData)context.Item).Result, (step) => step.UserFeedback)
	.Then<SayGoodbye>();`

How would you solve this?

PR #509, opened 2020-02-07
Overload the existing Setup() method in JsonWorkflowTest to wire up DI

coderookie1994:
Hi @danielgerlag 

While writing tests for my json based workflows, I noticed that there isn't any way of injecting mocks or any sort of dependency to the setup method. My workflow steps have multiple dependencies and while testing I want to wire up the mock objects.
To achieve that, I created a virtual overloaded method for the existing Setup() method, so users can pass IServiceCollection as input.
Please let me know if this change is correct or not.
Thanks

### DIFF
--- a/test/WorkflowCore.Testing/JsonWorkflowTest.cs
+++ b/test/WorkflowCore.Testing/JsonWorkflowTest.cs
@@ -39,6 +39,25 @@ namespace WorkflowCore.Testing
             Host.Start();
         }
 
+        protected virtual void Setup(IServiceCollection services)
+        {
+            // setup dependency injection
+            services.AddLogging();
+            ConfigureServices(services);
+
+            var serviceProvider = services.BuildServiceProvider();
+
+            //config logging
+            var loggerFactory = serviceProvider.GetService<ILoggerFactory>();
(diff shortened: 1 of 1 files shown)


test/WorkflowCore.Testing/JsonWorkflowTest.cs
```diff
@@ -39,6 +39,25 @@ protected virtual void Setup()
             Host.Start();
         }
 
+        protected virtual void Setup(IServiceCollection services)
```
danielgerlag: Could you not just override the `ConfigureServices` method in your tests?

coderookie1994: Yes, I see that I can override ConfigureServices, but the services variable declared in Setup() is a local variable. The Host is started using the local variable. I don't see a way by which I could pass my IServiceCollection, before the Host is started.

danielgerlag: You don't need to pass your own instance of IServiceCollection... you can interact with it in the `ConfigureServices` method

PR #524, opened 2020-02-24
Upgraded Nest to latest version of the package

julienperignon:


test/Docker.Testify/DockerSetup.cs
```diff
@@ -20,7 +20,7 @@ public abstract class DockerSetup : IDisposable
         public abstract int InternalPort { get; }
 
         public virtual string ImageTag => "latest";
-        public virtual TimeSpan TimeOut => TimeSpan.FromSeconds(30);
+        public virtual TimeSpan TimeOut => TimeSpan.FromSeconds(60);
```
danielgerlag: Could we just override this for elasticsearch, instead of everyone?

julienperignon: Totally can ! Will do it later today :)

PR #525, opened 2020-02-25
Try-catch clause

kirsanium:
Feature introducing complex error handling via new StepBuilder's methods Try() and Catch().
This works pretty similar to CompensateWith() and is based off of that, but with no reversion and it allows to catch exceptions by type. 
Also introducing WorkflowDefinitionValidator to ensure that there are corresponding catches for every try.


src/WorkflowCore/Interface/ICatchStepBuilder.cs
```diff
@@ -0,0 +1,9 @@
+﻿using System;
+
+namespace WorkflowCore.Interface
+{
+    public interface ICatchStepBuilder<TData, TStepBody> : IStepBuilder<TData, TStepBody>, ITryStepBuilder<TData, TStepBody>
```
danielgerlag: Do we need an empty interface?

src/WorkflowCore/Interface/IWorkflowDefinitionValidator.cs
```diff
@@ -0,0 +1,9 @@
+using WorkflowCore.Models;
+
+namespace WorkflowCore.Interface
+{
+    public interface IWorkflowDefinitionValidator
+    {
+        bool IsDefinitionValid(WorkflowDefinition definition);
```
danielgerlag: is this part of the try/catch feature? or something else?

src/WorkflowCore/Models/ExecutionPointer.cs
```diff
@@ -45,6 +45,8 @@ public class ExecutionPointer
         public object Outcome { get; set; }
 
         public PointerStatus Status { get; set; } = PointerStatus.Legacy;
+        
+        public Exception CurrentException { get; set; }
```
danielgerlag: This class needs to be serializable... not sure how different sub-classes of `Exception` will behave in that case?

src/WorkflowCore/Models/LifeCycleEvents/WorkflowTerminated.cs
```diff
@@ -1,10 +1,9 @@
 ﻿using System;
-using System.Collections.Generic;
-using System.Text;
 
 namespace WorkflowCore.Models.LifeCycleEvents
 {
     public class WorkflowTerminated : LifeCycleEvent
     {
+        public Exception Exception { get; set; }
```
danielgerlag: This class also needs to be serializable

src/WorkflowCore/Services/FluentBuilders/StepBuilder.cs
```diff
@@ -1,13 +1,16 @@
...
 using WorkflowCore.Interface;
 using WorkflowCore.Models;
 using WorkflowCore.Primitives;
 
 namespace WorkflowCore.Services
 {
-    public class StepBuilder<TData, TStepBody> : IStepBuilder<TData, TStepBody>, IContainerStepBuilder<TData, TStepBody, TStepBody>
+    public class StepBuilder<TData, TStepBody> 
+        : IContainerStepBuilder<TData, TStepBody, TStepBody>,
+          ICatchStepBuilder<TData, TStepBody>
```
danielgerlag: Not sure I understand why the class would implement ICatchStepBuilder?

src/WorkflowCore/Services/WorkflowRegistry.cs
```diff
@@ -11,10 +11,12 @@ public class WorkflowRegistry : IWorkflowRegistry
     {
         private readonly IServiceProvider _serviceProvider;
         private readonly List<Tuple<string, int, WorkflowDefinition>> _registry = new List<Tuple<string, int, WorkflowDefinition>>();
+        private readonly IWorkflowDefinitionValidator _validator;
 
-        public WorkflowRegistry(IServiceProvider serviceProvider)
+        public WorkflowRegistry(IServiceProvider serviceProvider, IWorkflowDefinitionValidator validator)
```
danielgerlag: Nice idea :+1:

src/WorkflowCore/Services/ErrorHandlers/CatchHandler.cs
```diff
@@ -0,0 +1,88 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using WorkflowCore.Interface;
+using WorkflowCore.Models;
+using WorkflowCore.Models.LifeCycleEvents;
+
+namespace WorkflowCore.Services.ErrorHandlers
+{
+    public class CatchHandler : IWorkflowErrorHandler
```
danielgerlag: Could you implement a set of unit tests for this class that make it's behavior explicit?

src/WorkflowCore/Interface/IWorkflowDefinitionValidator.cs
kirsanium: Right now this only checks that every try has corresponding catch.

src/WorkflowCore/Services/WorkflowRegistry.cs
kirsanium: I appreciate it!

src/WorkflowCore/Interface/ICatchStepBuilder.cs
kirsanium: This interface gives a builder full functionality of `IStepBuilder` and also allows to `Catch()` if we want to have multiple catches after one `Try()`. It is used as a return type for builder after the first `Catch()`, whereas after `Try()` we want to allow only to `Catch()`.

src/WorkflowCore/Services/FluentBuilders/StepBuilder.cs
kirsanium: Because we need to return `ICatchStepBuilder` after `Catch()`.
```
    public class StepBuilder<TData, TStepBody> 
        : IContainerStepBuilder<TData, TStepBody, TStepBody>,
          ICatchStepBuilder<TData, TStepBody>,
          IStepBuilder<TData, TStepBody>, 
          ITryStepBuilder<TData, TStepBody>
```
This would be more verbose, but two last lines are redundant in that case.

src/WorkflowCore/Models/ExecutionPointer.cs
kirsanium: As far as I understood, error handling happens in the same persistence round (not sure if this is how it's called tho) when the exception occurs. So do we really need to serialize this field?

src/WorkflowCore/Models/LifeCycleEvents/WorkflowTerminated.cs
kirsanium: Would it be fine if we would put `ExecutionError` in here instead?

src/WorkflowCore/Services/ErrorHandlers/CatchHandler.cs
kirsanium: Sure, will do it soon. By the way I have a question concerning behaviour of `CompensateHandler`. I don't understand what lines 41-45 of `CatchHandler` do as I copypasted them from lines 56-60 of `CompensateHandler`, hence I don't understand if they are needed at all. Could you please elaborate on this?

src/WorkflowCore/Models/ExecutionPointer.cs
danielgerlag: Why do we need this field on the `ExecutionPointer`?

src/WorkflowCore/Models/LifeCycleEvents/WorkflowTerminated.cs
danielgerlag: There already is ` WorkflowError` lifecycle event.  Also, you could terminate a workflow without an error... not sure this field makes sense?

src/WorkflowCore/Services/ErrorHandlers/CatchHandler.cs
danielgerlag: I believe that was for bubbling up the call stack of nested containers

danielgerlag:
Going to be travelling for a few weeks but I'll try take a deeper look soon.

kirsanium:
Just wanted to say I didn't forget about PR, going to fix the issues you noticed some time soon.

src/WorkflowCore/Models/ExecutionPointer.cs
kirsanium: @danielgerlag We need this field to build `StepExecutionContext` so that we can retrieve a thrown exception in `Catch()` step. An example:
```c#
builder.Try(b => b.StartWith(_ => throw new ArgumentException("I am Exception1")))
            .Catch(new[] {typeof(Exception)}, ctx =>
            {
                Console.WriteLine($"Caught an exception: Type: '{ctx.CurrentException.GetType().Name}', Message: '{ctx.CurrentException.Message}'");
                return ExecutionResult.Next();
            })
```

src/WorkflowCore/Models/LifeCycleEvents/WorkflowTerminated.cs
kirsanium: @danielgerlag 
When a workflow doesn't catch an exception thrown in `Try` clause, it seems logical to me that the workflow has to be terminated, because that means something has gone pretty badly. This can be the case if there is no matching exception type in corresponding `Catch` clauses. It means `WorkflowTerminated` event has to be published when this happens. I publish `WorkflowError` event when an exception happens inside of a `Try` clause, please, correct me, if I'm doing it wrong. If we want to retrieve information about why workflow has been terminated, we have to preserve an exception in some way. If a workflow was terminated without an error, this field just equals to `null`. 
Nevertheless, if this field needs to be serializable, I'm going to make it of a custom class with fields like Type, Message and StackTrace, because I feel like people might make a good use of the additional information on an unexpected workflow termination.

src/WorkflowCore/Services/ErrorHandlers/CatchHandler.cs
kirsanium: I clearly have troubles understanding this part of bubbling up the call stack. Could you please go into deeper details? And please tell me if it's needed or not in case of `Catch`clause.

danielgerlag:
@kirsanium I haven't forgotten about this PR.  I will try carve out some time in the next week or so to dive deeper into it but on the surface it looks really good!

knoxi:
@danielgerlag @kirsanium  Any news on this? Would be very handy to have this in the product.

kirsanium:
@knoxi waiting for the review from @danielgerlag

knoxi:
@danielgerlag  any news on that? It's now open since a year.

knoxi:
@kirsanium 
Do you see any chance to update the branch and extend the tests?
It would be really great if the feature will be part of the library!

kirsanium:
Sorry I have no time and interest in this one since too much time have passed - I haven't even coded in c# for some time now...but please please finish this one, it was goddamn ready two years ago

PR #576, opened 2020-06-08
Allow dependency injection on Workflow

glucaci:
It will be very usefull to can inject services/configuration into Workflow directly.

Example:
```csharp
public class SampleWorkflow : IWorkflow<SampleRequest>
{
    private readonly IOptions<WorkflowOptions> _options;

    public SampleWorkflow(IOptions<WorkflowOptions> options)
    {
        _options = options;
    }

    public string Id => nameof(SampleWorkflow);

    public int Version => 1;

    public void Build(IWorkflowBuilder<SampleRequest> builder)
    {
        builder
            .StartWith(context => ExecutionResult.Next())
            .Then<SampleStep>()
                .Input(step => step.Request, data => data)
                .OnError(WorkflowErrorHandling.Retry, _options.Value.RetryInterval);
    }
}
```

### DIFF
--- a/src/WorkflowCore/Interface/IWorkflowController.cs
+++ b/src/WorkflowCore/Interface/IWorkflowController.cs
@@ -13,8 +13,8 @@ namespace WorkflowCore.Interface
         Task<string> StartWorkflow<TData>(string workflowId, int? version, TData data = null, string reference=null) where TData : class, new();
 
         Task PublishEvent(string eventName, string eventKey, object eventData, DateTime? effectiveDate = null);
-        void RegisterWorkflow<TWorkflow>() where TWorkflow : IWorkflow, new();
-        void RegisterWorkflow<TWorkflow, TData>() where TWorkflow : IWorkflow<TData>, new() where TData : new();
+        void RegisterWorkflow<TWorkflow>() where TWorkflow : IWorkflow;
+        void RegisterWorkflow<TWorkflow, TData>() where TWorkflow : IWorkflow<TData> where TData : new();
 
         /// <summary>
         /// Suspend the execution of a given workflow until .ResumeWorkflow is called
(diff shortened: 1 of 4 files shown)


vaclavpfeifer:
I think it is already possible to register a WF using DI. 
RegisterWorkflow(serviceProvider.GetService<SomeDiRegisteredWf>());

glucaci:
Where is this method?
These are the only two methods of ```RegisterWorkflow```
```csharp
public interface IWorkflowController
{
    ...
    void RegisterWorkflow<TWorkflow>() where TWorkflow : IWorkflow, new();

    void RegisterWorkflow<TWorkflow, TData>()
      where TWorkflow : IWorkflow<TData>, new()
      where TData : new();
    ...
}
```

glucaci:
Ok, I saw... you mean on the ```IWorkflowRegistry```.
We used till now the IWorkflowHost to register the workflows because you need also to start it afterwards. It's pretty confusing registering workflows because this method is everywhere, IWorkflowHost , IWorkflowController, IWorkflowRegistry

glucaci:
The solution from ```IWorkflowRegistry``` implies that you have to ```new````manually the Workflow which is not very clean, because using the ```IServiceProvider``` is not an option

glucaci:
@danielgerlag what do you think? could we merge this?

glucaci:
Done.
Only Core right? Why are you not using same version for all the packages?

danielgerlag:
We use the same major version, but allow the minor and patch to vary.

IvanJosipovic:
@danielgerlag Do you have an ETA when this code will be released?

Thank you!

danielgerlag:
ah, it looks like it didn't deploy because the `PackageVersion` was not incremented, I'll sort that out!

PR #605, opened 2020-07-20
Steps after If should run when Schedule step is still Pending

andre-pt:
first of all, I'm not a .NET guy :)

we have been using jworkflow for a while now and recently a new use-case came up which in our opinion it's not the correct behaviour.

Consider the following workflow:
```
            builder
                .StartWith(context =>
                {
                    Console.WriteLine("STEP1");
                    return ExecutionResult.Next();
                })
                .Schedule(data => TimeSpan.FromSeconds(10)).Do(schedule => schedule
                    .StartWith(context => Console.WriteLine("STEP3"))
                )
                .If(data => 1 == 1).Do(then => then
                    .StartWith(context =>
                    {
                        Console.WriteLine("STEP2");
                        return ExecutionResult.Next();
                    })
                    .Then(Context =>
                    {
                        Console.WriteLine("STEP4");
                        return ExecutionResult.Next();
                    })
                )
                .Then(context =>
                {
                    Console.WriteLine("STEP5");
                    return ExecutionResult.Next();
                });
```

Our expected output would be:
```
STEP1
STEP2
STEP4
STEP5
STEP3
```

but it was actually
```
STEP1
STEP2
STEP4
STEP3
STEP5
```

During our investigation, we figured steps added after the If branch were never executed until the Schedule step was completed, so we have adapted the logic of `WorkflowCore.Services.WorkflowExecutor.DetermineNextExecutionTime`.

Is this a correct assumption or did we interpret the use case in a different way?

Thanks in advance.
Andre and @poberbichler

### DIFF
--- a/src/WorkflowCore/Services/WorkflowExecutor.cs
+++ b/src/WorkflowCore/Services/WorkflowExecutor.cs
@@ -211,25 +211,24 @@ namespace WorkflowCore.Services
             if (workflow.Status == WorkflowStatus.Complete)
                 return;
 
-            foreach (var pointer in workflow.ExecutionPointers.Where(x => x.Active && (x.Children ?? new List<string>()).Count == 0))
+            foreach (var pointer in workflow.ExecutionPointers.Where(x => x.Active))
             {
-                if (!pointer.SleepUntil.HasValue)
+                if ((pointer.Children ?? new List<string>()).Count == 0)
                 {
-                    workflow.NextExecution = 0;
-                    return;
+                    if (!pointer.SleepUntil.HasValue)
+                    {
+                        workflow.NextExecution = 0;
+                        return;
+                    }
+
+                    var pointerSleep = pointer.SleepUntil.Value.ToUniversalTime().Ticks;
+                    workflow.NextExecution = Math.Min(pointerSleep, workflow.NextExecution ?? pointerSleep);
                 }
-
-                var pointerSleep = pointer.SleepUntil.Value.ToUniversalTime().Ticks;
-                workflow.NextExecution = Math.Min(pointerSleep, workflow.NextExecution ?? pointerSleep);
-            }
-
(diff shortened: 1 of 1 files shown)


andre-pt:
hi @danielgerlag 
did you have an opportunity to look into this topic?

thanks
André

danielgerlag:
@andre-pt sorry this one slipped under my radar... I will try spend some time debugging & understanding the issue this week.

src/WorkflowCore/Services/WorkflowExecutor.cs
```diff
@@ -211,25 +211,24 @@ private void DetermineNextExecutionTime(WorkflowInstance workflow)
...
-                var pointerSleep = pointer.SleepUntil.Value.ToUniversalTime().Ticks;
-                workflow.NextExecution = Math.Min(pointerSleep, workflow.NextExecution ?? pointerSleep);
-            }
+                    if (!pointer.SleepUntil.HasValue)
+                    {
+                        workflow.NextExecution = 0;
+                        return;
+                    }
 
-            if (workflow.NextExecution == null)
```
danielgerlag, line 226: I think the only required change is to remove this `if` statement and always execute the inner code.
We can remove all the other changes... seems like an attempt for optimization that is causing a problem.

danielgerlag (changes requested):

src/WorkflowCore/Services/WorkflowExecutor.cs
danielgerlag: @andre-pt this is a valid bug that needs to be fixed.  Do you want to do it with this PR or should I open a separate one?

danielgerlag: https://github.com/danielgerlag/workflow-core/pull/700

Issue #607, opened 2020-07-22
Concurrency issues: The collection's state is no longer correct


andrx:
Using .net core 3.1 on unix, long running workflows.
Have been running the service on three nodes. two of them died (1st - after 5hrs, 2nd - after 8hrs) with the following exception.

	UnhandledExceptionTrapper - System.InvalidOperationException: Operations that change non-concurrent collections must have exclusive access. A concurrent update was performed on this collection and corrupted its state. The collection's state is no longer correct.
	   at System.Collections.Generic.HashSet`1.Contains(T item)
	   at WorkflowCore.Services.BackgroundTasks.QueueConsumer.<>c__DisplayClass15_1.<<Execute>b__0>d.MoveNext()
	   --- End of stack trace from previous location where exception was thrown ---
	      at System.Threading.Tasks.Task.<>c.<ThrowAsync>b__139_1(Object state)
	         at System.Threading.QueueUserWorkItemCallback.Execute()
	            at System.Threading.ThreadPoolWorkQueue.Dispatch()

Doesn't look like hard to fix. I can help. Do you prefer [ConcurrentHashSet nuget](https://github.com/i3arnon/ConcurrentHashSet) or just switching to ConcurrentDictionary? Another question would be why that happened?

danielgerlag:
Good catch... I believe this is happening because we are manipulating the (non-thread safe) set inside an async task which could run in another thread.  I would opt to avoid adding more library dependencies if possible to keep this light-weight.

andrx:
I can confirm switching to ConcurrentHashSet fixed the issue. 3 days under load - didn't happen anymore

Issue #414, opened 2019-10-04
Getting an IndexOutOfRangeException on QueueConsumer 


siskender:
Hi,

Since yesterday (after a new deployment) of our workflow component were we are using this workflow core component we are getting the following exception:

```json
{"@t":"2019-10-04T11:45:07.7109233Z","@m":"Index was outside the bounds of the array.","@i":"da24d145","@l":"Error","@x":"System.IndexOutOfRangeException: Index was outside the bounds of the array.\n   at System.Collections.Generic.HashSet`1.SetCapacity(Int32 newSize)\n   at System.Collections.Generic.HashSet`1.IncreaseCapacity()\n   at System.Collections.Generic.HashSet`1.AddIfNotPresent(T value)\n   at WorkflowCore.Services.BackgroundTasks.QueueConsumer.Execute()","SourceContext":"WorkflowCore.Services.BackgroundTasks.IndexConsumer","ExceptionDetail":{"Type":"System.IndexOutOfRangeException","HResult":-2146233080,"Message":"Index was outside the bounds of the array.","Source":"System.Collections"},"Stack":{"StackName":"******","Environment":"******","$type":"StackOptions"}}
```
Investigating it further I can pinpoint that in the QueueConsumer there is a HashSet for the second passes. But my question is what is the purpose of this and why do you think that this is giving an error.

For the QueueProvider we are using Redis.

danielgerlag:
This is high priority, I will investigate this ASAP

danielgerlag:
I can't reproduce this issue... are you able to send a sample project that replicates the issue?
Which version are you using? Could you describe your environment in more detail?

siskender:
Hi,

It is very difficult to give an example (because of sensitivity and complexity). But I can describe it for you.

We have 2 main workflows running, that can conditionally trigger other workflows (as subworkflows). These subworkflows, when they are finished, send events to the main workflows and close them self. The main workflows will then continue doing there stuff. But we have a lot of concurrent workflows running at the same time (either main ones, either sub ones). These can stay open for several days before getting finished (with a WaitFor).

The stack we are using:
- .NET Core 2.2
- Workflow Core 2.0
- Redis for locking and events
- Mongo for persistence

We are running the workflows for months on environments outside of production and we never saw this error appearing. We saw this error appearing in production after doing an update to an existing running instance of the workflow.

siskender:
An update on our side for this:
[...]

PR #608, opened 2020-07-23
#607 switched to ConcurrentHashSet.

andrx:


siskender:
This should solve the following Exception we are occasionally getting in a multi-service (multi-threaded) environment.

System.IndexOutOfRangeException: Index was outside the bounds of the array.
at System.Collections.Generic.HashSet`1.SetCapacity(Int32 newSize)
at System.Collections.Generic.HashSet`1.AddIfNotPresent(T value)
at WorkflowCore.Services.BackgroundTasks.QueueConsumer.Execute()

https://stackoverflow.com/questions/4307131/indexoutofrangeexception-when-adding-to-hashsett

Let's be sure that this PR also resolves this issue

siskender:
I also created an issue for this https://github.com/danielgerlag/workflow-core/issues/414

andrx:
yes, this change worked for me as well but @danielgerlag asked to switch to `ConcurrentDictionary`. Just need a moment to update the PR.

Issue #606, opened 2020-07-21
Polling Interval. Strange behaviour.


andrx:
I have 3 instances of workflow-core hosts running (.net core 3.1) just one workflow definition as a proof of concept. To minimize the number of new techs decided to use redis as it has providers for all layers: locking queuing, persistence, events. After a week running noticed that redis performance started degrading.  Bigkeys command showed that wfc keys consumed most of the resources.
Ok, bad choice obviously. Quick changes applied to move persistence to mongodb and queuing to rabbitmq. After restart noticed that wcf.workflow_queue started growing and growing and never stopped though no new workflows started. I killed the instances and removed the queues after it consumed several GBs on rabbitmq. Mongodb had 50k docs. Checked the code and looks like the only process responsible for this lives in the [Poller](https://github.com/danielgerlag/workflow-core/blob/master/src/WorkflowCore/Services/BackgroundTasks/RunnablePoller.cs
) and by default every 10 sec it locks and pulls all runnable instances from persistence layer and pushes those to the above mentioned queue. It does the same for events but I don’t have any events just yet. 

Then I changed polling interval to 2min and restarted. It didn’t help. The incoming speed for that queue was as twice as higher than reading. Then I tried with 5 minutes interval. That helped. It was a huge bump in the beginning but by the time next polling happened the queue was empty. Incoming dropped. Delivery was on the same level and then these two were equal.

This is the part I don’t get and why that happened. I can read the code but would like to understand the logic and what I’m doing wrong. Maybe the way I use it and my expectations are not correct.

Today I tried with 30sec polling interval. Obviously 5min – is too long.
And this is what I saw after restart.
### **First 10 mins:**
![wcf.workflow_queue](https://files.slack.com/files-pri/T0HNY7Z99-F017G1P3CUS/image.png?pub_secret=e7afb9cea5)

### **20mins:**
![wcf.workflow_queue](https://files.slack.com/files-pri/T0HNY7Z99-F0171AMSCRM/image.png?pub_secret=90252123a4)
![wcf.index_queue](https://files.slack.com/files-pri/T0HNY7Z99-F017N8G2WKE/image.png?pub_secret=9af323d490)

### **30mins:**
![wcf.index_queue](https://files.slack.com/files-pri/T0HNY7Z99-F017G3P3V18/image.png?pub_secret=8127329526)
**also went down**

So, it eventually slowed down and kept in a normal pace even starting new worklflow instances.

If you could suggest anything.. what is wrong.

[...]

PR #632, opened 2020-08-19
#606 switched activeTasks to ConcurrenDictionary

andrx:
Switched to `activeTasks` to ConcurrentDictionary
Added logging about the number of events polled. 

Will be load testing it for a few more days.

Also not sure what `greyList` is (recent updates in master)?

### DIFF
--- a/src/WorkflowCore/Services/BackgroundTasks/RunnablePoller.cs
+++ b/src/WorkflowCore/Services/BackgroundTasks/RunnablePoller.cs
@@ -54,7 +54,8 @@ namespace WorkflowCore.Services.BackgroundTasks
                     try
                     {
                         _logger.LogInformation("Polling for runnable workflows");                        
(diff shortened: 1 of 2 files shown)


src/WorkflowCore/Services/BackgroundTasks/RunnablePoller.cs
```diff
@@ -54,7 +54,8 @@ private async void PollRunnables(object target)
                     try
                     {
                         _logger.LogInformation("Polling for runnable workflows");                        
-                        var runnables = await _persistenceStore.GetRunnableInstances(DateTime.Now);
+                        var runnables = (await _persistenceStore.GetRunnableInstances(DateTime.Now)).ToArray();
+                        _logger.LogInformation($"Polled {runnables.Length} runnable workflows");
```
andrx, line 58: Probably would make more sense to just leave the log with the number of polled records...

PR #637, opened 2020-08-31
Fix #595 for Queue Scheduling.

### LINKED ISSUE #595: Question: Performance and Queuing Mechanism
I have a question regarding workflow throughput and how it relates to the queuing mechanism.

**Workflow Overview**
1. Download a document.
2. Send document to preprocessing service via an API call
3. Wait for an event (which will be emitted when preprocessing service has completed request async)
4. Send preprocessed data to machine learning inference service via an API call
5. Wait for an event (which will be emitted when the ML service has completed request async)
6. Hit a callback API with the results of the workflow.

I am using the Redis provider for queuing and locking, though it should be noted that only a single node is running for this test.

**Performance Results**
If I execute a single request, the entire process takes about 3.6 seconds, which is great. I started sending more and more requests through the application to see how responses work at scale. The time for the _first_ request to respond scaled essentially linearly with the number of requests. So if I send through 10 requests, the first one won't complete until 4.7 seconds. At 100 requests, 25.12 seconds. At 500, 107.35.

I have included a graph of the min, max and avg responses time as I've increased the number of jobs below.

image

What this seems to suggest to me is that steps are more or less being processed sequentially across all workflows. Meaning for request 1, we don't process step 3 until all step 2 workflow requests have completed. I can see why that might be the case, particularly since Redis queuing is a simple list structure.

Given that steps 2 and 4 are asynchronous, external APIs, we could see a massive increase in processing time if we could take advantage of that information. Do you have any recommendations on how to achieve something like this?

Please let me know if there is any more information I can provide for you.

ccavin:
Executes all ExecutionPointers for a workflow including those generated as a
result of a previous ExecutionPointer.

### DIFF
--- a/src/WorkflowCore/Services/WorkflowExecutor.cs
+++ b/src/WorkflowCore/Services/WorkflowExecutor.cs
@@ -41,7 +41,6 @@ namespace WorkflowCore.Services
         {
             var wfResult = new WorkflowExecutorResult();
 
-            var exePointers = new List<ExecutionPointer>(workflow.ExecutionPointers.Where(x => x.Active && (!x.SleepUntil.HasValue || x.SleepUntil < _datetimeProvider.UtcNow)));
             var def = _registry.GetDefinition(workflow.WorkflowDefinitionId, workflow.Version);
             if (def == null)
             {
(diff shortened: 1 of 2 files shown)


ccavin:
This appears to completely break compensating transactions and some of the cancellation unit tests.

PR #655, opened 2020-09-25
Feature/rabbit mq multinode connection

shirados90:
My team has encountered with problems of multi-node RabbitMq connection due to RabbitMq.Client specific connection API.
These changes just extend RabbitMqQueueProvider configuration

Also there's a dedicated IRabbitMqQueueNameProvider allowing to hook into queueNaming from outside

### DIFF
--- a/src/providers/WorkflowCore.QueueProviders.RabbitMQ/ServiceCollectionExtensions.cs
+++ b/src/providers/WorkflowCore.QueueProviders.RabbitMQ/ServiceCollectionExtensions.cs
@@ -41,13 +41,15 @@ namespace Microsoft.Extensions.DependencyInjection
             if (rabbitMqConnectionFactory == null) throw new ArgumentNullException(nameof(rabbitMqConnectionFactory));
 
             options.Services.AddSingleton(rabbitMqConnectionFactory);
-            options.Services.TryAddTransient<IRabbitMqQueueNameProvider, DefaultRabbitMqQueueNameProvider>();
+            options.Services.TryAddSingleton<IRabbitMqQueueNameProvider, DefaultRabbitMqQueueNameProvider>();
             options.UseQueueProvider(RabbitMqQueueProviderFactory);
             
             return options;
         }
 
         private static IQueueProvider RabbitMqQueueProviderFactory(IServiceProvider sp)
-            => new RabbitMQProvider(sp);
+            => new RabbitMQProvider(sp,
+                sp.GetRequiredService<IRabbitMqQueueNameProvider>(),
+                sp.GetRequiredService<RabbitMqConnectionFactory>());
     }
 }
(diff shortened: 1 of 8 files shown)


src/providers/WorkflowCore.QueueProviders.RabbitMQ/Services/RabbitMQProvider.cs
```diff
@@ -7,23 +7,30 @@
...
+        private readonly RabbitMqConnectionFactory _rabbitMqConnectionFactory;
+        private readonly IServiceProvider _serviceProvider;
+        
         private IConnection _connection = null;
         private static JsonSerializerSettings SerializerSettings = new JsonSerializerSettings() { TypeNameHandling = TypeNameHandling.All };
 
         public bool IsDequeueBlocking => false;
 
-        public RabbitMQProvider(IConnectionFactory connectionFactory)
+        public RabbitMQProvider(IServiceProvider serviceProvider)
```
danielgerlag, line 29: I think we should inject the actual dependencies rather than the IoC container?

src/providers/WorkflowCore.QueueProviders.RabbitMQ/ServiceCollectionExtensions.cs
```diff
@@ -2,18 +2,52 @@
...
+                .UseRabbitMQ((sp, name) => connectionFactory.CreateConnection(hostnames.ToList(), name));
+        }
+        
+        public static WorkflowOptions UseRabbitMQ(this WorkflowOptions options, RabbitMqConnectionFactory rabbitMqConnectionFactory)
+        {
+            if (options == null) throw new ArgumentNullException(nameof(options));
+            if (rabbitMqConnectionFactory == null) throw new ArgumentNullException(nameof(rabbitMqConnectionFactory));
+
+            options.Services.AddSingleton(rabbitMqConnectionFactory);
+            options.Services.TryAddTransient<IRabbitMqQueueNameProvider, DefaultRabbitMqQueueNameProvider>();
```
danielgerlag, line 44: should we make thins singleton too?

danielgerlag (commented):
Thanks for your contribution.  Could you also increment the versions of the affected libraries?

src/providers/WorkflowCore.QueueProviders.RabbitMQ/ServiceCollectionExtensions.cs
shirados90: Changed to singleton

src/providers/WorkflowCore.QueueProviders.RabbitMQ/Services/RabbitMQProvider.cs
shirados90: Explicit dependencies added

shirados90:
> Thanks for your contribution. Could you also increment the versions of the affected libraries?

I've incremented minor version

danielgerlag (approved):

PR #662, opened 2020-10-01
#660 Don't lose context in When(_ => 0).Do(

### LINKED ISSUE #660: When causes context.Item to lose it's value
If you use  When(_ => 0)  instead of When(0) your context.Item loses it's value in the sub tasks.
I'm using it inside a ForEach-step.

When using when with expressions ,context.Item is null
```
.When(_ => 0).Do(then => 
   then.StartWith<DummyStep>()
      .Input(step => step.Value, (_, context) => context.Item)
```

The [Obsolete]-tagged When works fine, a.k.a context.Item is not null
```
.When(0).Then<DummyStep>()
   .Input(step => step.Value, (_, context) => context.Item)
```

ellinge:
A possible solution. Not sure this is good but you do a similar solution for outcome in the When-primitive.

And the pointer on the top level should be able to extract it's context from the children as well.

The outcome in at least becomes, which would fix #660:
> When(true/false): Approver = User 2 (1 executions) context.Item = 1
> When(true/false): Approver = User 1 (1 executions) context.Item = 0
> When(true/false): Approver = User 2 (2 executions) context.Item = 1
> When(true/false): Approver = User 1 (2 executions) context.Item = 0
> When(true): Approver = User 2 (3 executions) context.Item = 1
> When(true): Approver = User 1 (3 executions) context.Item = 0
> When(_ > true/false): Approver = User 2 (1 executions) context.Item = 1
> When(_ > true/false): Approver = User 1 (1 executions) context.Item = 0
> When(_ > true/false): Approver = User 2 (2 executions) context.Item = 1
> When(_ > true/false): Approver = User 1 (2 executions) context.Item = 0
> When(_ > true): Approver = User 2 (3 executions) context.Item = 1
> When(_ > true): Approver = User 1 (3 executions) context.Item = 0

### DIFF
--- a/src/WorkflowCore/Services/WorkflowExecutor.cs
+++ b/src/WorkflowCore/Services/WorkflowExecutor.cs
@@ -138,6 +138,7 @@ namespace WorkflowCore.Services
 
         private async Task ExecuteStep(WorkflowInstance workflow, WorkflowStep step, ExecutionPointer pointer, WorkflowExecutorResult wfResult, WorkflowDefinition def, CancellationToken cancellationToken = default)
         {
+            pointer.ContextItem = pointer.ContextItem ?? workflow?.ExecutionPointers?.FirstOrDefault(x => x.Children.Contains(pointer.Id))?.ContextItem;
             IStepExecutionContext context = new StepExecutionContext()
             {
                 Workflow = workflow,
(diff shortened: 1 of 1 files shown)


danielgerlag:
Closing in favor of https://github.com/danielgerlag/workflow-core/pull/683

PR #681, opened 2020-10-20
add support for WorkflowStep

wangshuai-007:
Before, the `StepType` field  must be subclass of `StepBody`
Now,It support for `WorkflowStep`
Old:
```c#
public class HelloWorld : StepBody
```
new:(Add)
```c#
public class EndStep : WorkflowStep
```
It means , In the `YourUserTaskStep`
you can set `ValueOutcome.Value` and control step create in [ProcessExecutionResult](https://github.com/danielgerlag/workflow-core/blob/32fedfd528e9d8f8eeae7a3b9500d84c5c411c39/src/WorkflowCore/Services/ExecutionResultProcessor.cs#L65)

### DIFF
--- a/src/WorkflowCore.DSL/Services/DefinitionLoader.cs
+++ b/src/WorkflowCore.DSL/Services/DefinitionLoader.cs
@@ -68,8 +68,22 @@ namespace WorkflowCore.Services.DefinitionStorage
                 var nextStep = stack.Pop();
 
                 var stepType = FindType(nextStep.StepType);
-                var containerType = typeof(WorkflowStep<>).MakeGenericType(stepType);
-                var targetStep = (containerType.GetConstructor(new Type[] { }).Invoke(null) as WorkflowStep);
+
+                WorkflowStep targetStep;
+
+                Type containerType;
+                if (stepType.IsSubclassOf(typeof(StepBody)))
+                {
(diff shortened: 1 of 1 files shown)


src/WorkflowCore.DSL/Services/DefinitionLoader.cs
```diff
@@ -68,8 +68,22 @@ private WorkflowStepCollection ConvertSteps(ICollection<StepSourceV1> source, Ty
                 var nextStep = stack.Pop();
 
                 var stepType = FindType(nextStep.StepType);
-                var containerType = typeof(WorkflowStep<>).MakeGenericType(stepType);
-                var targetStep = (containerType.GetConstructor(new Type[] { }).Invoke(null) as WorkflowStep);
+
+                WorkflowStep targetStep;
+
+                Type containerType;
+                if (stepType.IsSubclassOf(typeof(StepBody)))
```
jdevillard, line 75: Hello, 

Not sure you will handle all step with typeof(StepBody). (missing StepBodyAsync)
Would it not be better to use stepType.GetInterfaces().Contains(typeof(IStepBody)) ?

wangshuai-007: Yes,you are right.I missed `StepBodyAsync`

danielgerlag (approved):

PR #695, opened 2020-11-12
MIGRATION app csproj to net5.0, updated all project references

digitalsigi:
CHANGE SQL Provider uses now UseIdentityColumn
CHANGE some fluentassertions seem to be no longer there, changed the respective sources

### DIFF
--- a/src/providers/WorkflowCore.Persistence.PostgreSQL/WorkflowCore.Persistence.PostgreSQL.csproj
+++ b/src/providers/WorkflowCore.Persistence.PostgreSQL/WorkflowCore.Persistence.PostgreSQL.csproj
@@ -3,7 +3,7 @@
   <PropertyGroup>
     <AssemblyTitle>Workflow Core PostgreSQL Persistence Provider</AssemblyTitle>
     <Authors>Daniel Gerlag</Authors>
-    <TargetFramework>netstandard2.1</TargetFramework>
+    <TargetFramework>net5.0</TargetFramework>
     <AssemblyName>WorkflowCore.Persistence.PostgreSQL</AssemblyName>
     <PackageId>WorkflowCore.Persistence.PostgreSQL</PackageId>
     <PackageTags>workflow;.NET;Core;state machine;WorkflowCore;PostgreSQL</PackageTags>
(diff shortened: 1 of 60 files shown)


digitalsigi:
I did not update all version strings, because I had no glue what your numbering schema would be.  
However I updated those for WorkflowCore, workflowcore.persistence.entityframework and workflowcore.persistence.sqlserver

src/providers/WorkflowCore.Persistence.PostgreSQL/WorkflowCore.Persistence.PostgreSQL.csproj
```diff
@@ -27,12 +27,12 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Npgsql" Version="4.1.3.1" />
-    <PackageReference Include="Npgsql.EntityFrameworkCore.PostgreSQL" Version="3.1.2" />
-    <PackageReference Include="Microsoft.EntityFrameworkCore.Tools" Version="3.1.2">
+    <PackageReference Include="Npgsql" Version="4.1.5" />
+    <PackageReference Include="Npgsql.EntityFrameworkCore.PostgreSQL" Version="3.1.4" />
```
evolvedlight, line 31: can you bump these to v5 please?
https://www.nuget.org/packages/Npgsql.EntityFrameworkCore.PostgreSQL/
https://www.nuget.org/packages/Npgsql/

danielgerlag (commented):
As this is a library, I would prefer not to force all consumers of the library to upgrade in-step.  I would prefer to target the broadest possible set of framework versions out there, unless there is a specific feature in the newer version that we require.

evolvedlight:
@danielgerlag EF Core 5.0 target .Net Standard 2.1, so isn't compatible with .Net Framework anymore. 
But if you don't upgrade to EF Core 5.0, then we can't use this library in a .Net 5.0/Ef Core 5.0 application, as it's no longer compatible. So it seems for workflow core you'll need to choose between either supporting ef core 5.0, or supporting .Net Framework.
For us this is a problem - we'd like to use EF Core 5.0 features in an application that also has workflow core dependency - so at the moment workflow core is stopping us upgrading.

digitalsigi:
@danielgerlag Same here, had the requirement to upgrade to .net50 so I updated WorkflowCore. I am aware that this might result in some discrepances, but saw no other choice.

digitalsigi:
@evolvedlight @danielgerlag I will investigate whether a multi-target approach is possible. But that may take time as I have no experience in this corner.

danielgerlag:
In this case, we should only upgrade the libraries that depend  on EF core, rather than everything.

evolvedlight:
@danielgerlag to be clear, does that mean you will accept a PR that means that WorkflowCore looses support for .Net Framework?

danielgerlag:
@evolvedlight no, not for the core library.  We can update the EF Core libraries but keep all others targeting .Net Standard 2.0

evolvedlight:
@danielgerlag Sorry, of course. I'll make a PR for this - would you like me to bump the major version of the libraries too?

glucaci:
I would also don't do any change to core libraries, it should stay as low as possible with the target framework, than you don't have to change the version.

digitalsigi:
Fork deleted. Thx to all.

Issue #709, opened 2020-12-01
Middleware Scope Causes Threading Issues When Running Multiple Workflows


dflor003:
The way that the middlewares are scoped in #678 seems to be causing threading issues when used alongside the EntityFramework core persistence provider. Since the middleware executor is singleton scoped, that causes middleware that use a `DbContext` to effectively be singleton scoped even if they are declared as transient.

I'm going to work on a PR to fix this and move the resolution of the `StepExecutor` and other related classes into a scoped block right before use. Any objections?

dflor003:
Fixed by #710

Issue #678, opened 2020-10-18
Feature Proposal - Workflow and Step Middleware


dflor003:
## Proposal

My team has been experimenting with this library for a couple of projects and so far it's been quite a pleasure to work with, so thank you very much for that!

One area, though, where I see an opportunity for improvement around extensibility for various use-cases within the consuming apps. I propose adding a few additional hooks at the workflow and step level in the form of middleware to enable more extensibility for just about any use-case. Please take a look at the proposal below and if everything makes sense, I'd be happy to put together a PR to add this functionality.

These middleware would be similar in form to the type of middleware you would find around the [ASP.NET Core Controller middleware](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/middleware/write?view=aspnetcore-3.1) or [`HttpClient`'s `DelegatingHandler` middleware](https://docs.microsoft.com/en-us/aspnet/web-api/overview/advanced/httpclient-message-handlers) so it should feel familiar to enough to dotnet core users.

Here's an example of a few use-cases that this could help us solve:

* More robust retry strategies such as circuit breakers, exponential retry, and fallback using other libraries such as [Polly](https://github.com/App-vNext/Polly). There are a few existing issues that this may address: #654, #508 
* Injecting log correlation information into each workflow step so that we can view logs by workflow id.
* Appending additional metadata to workflow steps to support UIs. This would help with these issues: #672, #665
* Provide a general, unobtrusive hook into workflows and steps to make it easier to implement additional functionality without having to code for it in every step. 

## Examples

A line of code is worth a million pictures, so here are a few examples of how I think these middleware could look from the consumer perspective.

### Example 1 - Use a step middleware to add log correlation data to workflow steps

This is an example of how a middleware would look to add the workflow ID to log context using the [Serilog](https://serilog.net/) logging library and their built-in [logging context](https://nblumhardt.com/2016/08/context-and-correlation-structured-logging-concepts-in-net-5/#ambient-logging-contexts).

```cs
using Serilog;

public class SerilogCorrelationMiddleware : IWorkflowStepMiddleware {
  private const string WorkflowIdContextProperty = "workflowId";
  private readonly ILogger _log;

  public SerilogCorrelationMiddleware(ILogger logger) {
    _log = logger.ForContext<SerilogCorrelationMiddleware>();
  }

  public async Task<ExecutionResult> HandleAsync(
    IStepExecutionContext context, 
    IStepBody body, 
    WorkflowStepDelegate next
[...]

PR #710, opened 2020-12-01
Feature - Scoped Middleware

dflor003:
Move WorkflowMiddlewareRunner and StepExecutor to be instantiated in a scope specific to the step to avoid potential threading issues when used with EFCore and DbContext.

Fixes bug introduced in #678 and closes issue #709.

### DIFF
--- a/src/WorkflowCore/Services/WorkflowController.cs
+++ b/src/WorkflowCore/Services/WorkflowController.cs
@@ -21,10 +21,9 @@ namespace WorkflowCore.Services
         private readonly IExecutionPointerFactory _pointerFactory;
         private readonly ILifeCycleEventHub _eventHub;
         private readonly IServiceProvider _serviceProvider;
-        private readonly IWorkflowMiddlewareRunner _middlewareRunner;
         private readonly ILogger _logger;
 
-        public WorkflowController(IPersistenceProvider persistenceStore, IDistributedLockProvider lockProvider, IWorkflowRegistry registry, IQueueProvider queueProvider, IExecutionPointerFactory pointerFactory, ILifeCycleEventHub eventHub, ILoggerFactory loggerFactory, IServiceProvider serviceProvider, IWorkflowMiddlewareRunner middlewareRunner)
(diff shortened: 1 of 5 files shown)


src/WorkflowCore/Services/WorkflowController.cs
```diff
@@ -85,7 +83,11 @@ public async Task<string> StartWorkflow<TData>(string workflowId, int? version,
 
             wf.ExecutionPointers.Add(_pointerFactory.BuildGenesisPointer(def));
 
-            await _middlewareRunner.RunPreMiddleware(wf, def);
+            using (var scope = _serviceProvider.CreateScope())
```
danielgerlag, line 86: Should we use `IScopeProvider` instead?

src/WorkflowCore/Services/WorkflowExecutor.cs
```diff
@@ -250,7 +247,11 @@ private async Task DetermineNextExecutionTime(WorkflowInstance workflow, Workflo
             workflow.Status = WorkflowStatus.Complete;
             workflow.CompleteTime = _datetimeProvider.UtcNow;
 
-            await _middlewareRunner.RunPostMiddleware(workflow, def);
+            using (var scope = _serviceProvider.CreateScope())
```
danielgerlag, line 250: Should we use `IScopeProvider` instead?

src/WorkflowCore/Services/WorkflowController.cs
dflor003: I think it wouldn't be possible here because the scope provider relies on passing an `IStepExecutionContext`

src/WorkflowCore/Services/WorkflowExecutor.cs
dflor003: Same here, since we do not have an `IStepExecutionContext`.

danielgerlag (approved):
Could you increment the patch version?

dflor003:
Done! Incremented to 3.3.3.0

PR #713, opened 2020-12-03
Tackle consumer starvation

glucaci:
It will be nice to control which background task is starting. In my case i want to separate the workflow poller from workflow consumer.

Setup:
- n x Workflow Producers (this can be achieved by not starting the host and just register the workflow)
- 1 x Workflow Poller (i want to be able to start the host without to have the consumer enabled)
- n x Workflow Consumers (i want to be able to start the host without to have the poller enabled)

The simple way is to add option for each background worker to can disable them and don't register it in DI container.

If you have another idea or thought about this please let me know.

Thanks

**Update:**
Short description of the problem: We are running the workflow host inside the application host and this is deployed in multiple instances. With  enough amount of runnable workflows and keeping the polling interval to 10s you bring consumers to starvations. That is why we wanted to have an option to deactivat the poller.

But know i found another solution that we can replace the `IGreyList` with a async ready variant and implement this in `MongoDB` or `Redis`

Let me know what you think.

### DIFF
--- a/src/WorkflowCore/Models/WorkflowOptions.cs
+++ b/src/WorkflowCore/Models/WorkflowOptions.cs
@@ -34,6 +34,11 @@ namespace WorkflowCore.Models
             EventHubFactory = new Func<IServiceProvider, ILifeCycleEventHub>(sp => new SingleNodeEventHub(sp.GetService<ILoggerFactory>()));
         }
 
+        public bool DisableWorkflowConsumer { get; set; } = false;
+        public bool DisableEventConsumer { get; set; } = false;
+        public bool DisableIndexConsumer { get; set; } = false;
+        public bool DisableRunnablePoller { get; set; } = false;
+
         public void UsePersistence(Func<IServiceProvider, IPersistenceProvider> factory)
         {
             PersistanceFactory = factory;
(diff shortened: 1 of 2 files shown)


src/WorkflowCore/Models/WorkflowOptions.cs
```diff
@@ -34,6 +34,11 @@ public WorkflowOptions(IServiceCollection services)
             EventHubFactory = new Func<IServiceProvider, ILifeCycleEventHub>(sp => new SingleNodeEventHub(sp.GetService<ILoggerFactory>()));
         }
 
+        public bool DisableWorkflowConsumer { get; set; } = false;
+        public bool DisableEventConsumer { get; set; } = false;
+        public bool DisableIndexConsumer { get; set; } = false;
+        public bool DisableRunnablePoller { get; set; } = false;
```
danielgerlag, line 40: What do you think of flipping the naming here and default to true... something like `ProcessWorkflows`, `ProcessEvents`, `PollingEnabled` or some variant?

danielgerlag (approved):

src/WorkflowCore/Models/WorkflowOptions.cs
glucaci: Yes, no problem

glucaci:
Done. I'll do another PR for `IGreyList`

danielgerlag:
@glucaci thanks... looks like the build is failing though

glucaci:
@danielgerlag done

PR #715, opened 2020-12-04
Add dotnet 5 support (with fixed mysql)

evolvedlight:
Main tests all now working:

image

### DIFF
--- a/src/WorkflowCore/WorkflowCore.csproj
+++ b/src/WorkflowCore/WorkflowCore.csproj
@@ -15,12 +15,12 @@
     <GenerateAssemblyCompanyAttribute>false</GenerateAssemblyCompanyAttribute>
     <GenerateAssemblyProductAttribute>false</GenerateAssemblyProductAttribute>
     <Description>Workflow Core is a light weight workflow engine targeting .NET Standard.</Description>
-    <Version>3.3.2</Version>
-    <AssemblyVersion>3.3.2.0</AssemblyVersion>
-    <FileVersion>3.3.2.0</FileVersion>
+    <Version>3.3.3</Version>
+    <AssemblyVersion>3.3.3.0</AssemblyVersion>
+    <FileVersion>3.3.3.0</FileVersion>
     <PackageReleaseNotes></PackageReleaseNotes>
     <PackageIconUrl>https://github.com/danielgerlag/workflow-core/raw/master/src/logo.png</PackageIconUrl>
-    <PackageVersion>3.3.2</PackageVersion>
+    <PackageVersion>3.3.3</PackageVersion>
   </PropertyGroup>
 
   <ItemGroup>
--- a/src/providers/WorkflowCore.Persistence.EntityFramework/WorkflowCore.Persistence.EntityFramework.csproj
+++ b/src/providers/WorkflowCore.Persistence.EntityFramework/WorkflowCore.Persistence.EntityFramework.csproj
@@ -3,7 +3,7 @@
   <PropertyGroup>
     <AssemblyTitle>Workflow Core EntityFramework Core Persistence Provider</AssemblyTitle>
     <Authors>Daniel Gerlag</Authors>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFramework>netstandard2.1</TargetFramework>
     <AssemblyName>WorkflowCore.Persistence.EntityFramework</AssemblyName>
     <PackageId>WorkflowCore.Persistence.EntityFramework</PackageId>
     <PackageTags>workflow;.NET;Core;state machine;WorkflowCore;EntityFramework;EntityFrameworkCore</PackageTags>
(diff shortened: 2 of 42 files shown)


src/providers/WorkflowCore.Providers.Redis/WorkflowCore.Providers.Redis.csproj
```diff
@@ -2,18 +2,18 @@
...
     <RepositoryType>git</RepositoryType>
     <PackageProjectUrl>https://github.com/danielgerlag/workflow-core</PackageProjectUrl>
     <Description>Redis providers for Workflow Core (Persistence, queueing, distributed locking and event hubs)
 </Description>
-    <PackageVersion>3.0.2</PackageVersion>
+    <PackageVersion>3.0.3</PackageVersion>
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="2.2.0" />
```
danielgerlag, line 16: Do we need to update the redis package to support the EF upgrade?

src/providers/WorkflowCore.Providers.Elasticsearch/WorkflowCore.Providers.Elasticsearch.csproj
```diff
@@ -13,7 +13,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="2.2.0" />
```
danielgerlag, line 16: Do we need to update the Elastic package to support the EF upgrade?

src/providers/WorkflowCore.Providers.AWS/WorkflowCore.Providers.AWS.csproj
```diff
@@ -11,16 +11,16 @@
...
+    <Version>3.0.3</Version>
+    <AssemblyVersion>3.0.3.0</AssemblyVersion>
+    <PackageVersion>3.0.3</PackageVersion>
   </PropertyGroup>
 
   <ItemGroup>
     <PackageReference Include="AWSSDK.DynamoDBv2" Version="3.3.104.7" />
     <PackageReference Include="AWSSDK.Kinesis" Version="3.3.100.112" />
     <PackageReference Include="AWSSDK.SQS" Version="3.3.102.44" />
-    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="2.2.0" />
```
danielgerlag, line 23: Do we need to update the AWS package to support the EF upgrade?

evolvedlight:
I’ll check these comments on Monday, travelling today

src/providers/WorkflowCore.Providers.Redis/WorkflowCore.Providers.Redis.csproj
evolvedlight: It might be that it runs ok with mixed versions, but EF Core 5.0 requires Microsoft.Extensions.Logging => 5.0.0 which requires the abstractions package => 5.0.0.

I don't think this is a difficult upgrade for clients to use 5.0.0 instead of 2.2.0, and it seems better than leaving the redis package on a different version to the EF Core related libs?

src/WorkflowCore/WorkflowCore.csproj
```diff
@@ -15,20 +15,20 @@
...
   <ItemGroup>
     <PackageReference Include="ConcurrentHashSet" Version="1.1.0" />
-    <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="2.2.0" />
-    <PackageReference Include="Microsoft.Extensions.Hosting.Abstractions" Version="2.2.0" />
-    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="2.2.0" />
-    <PackageReference Include="Microsoft.Extensions.ObjectPool" Version="2.2.0" />
+    <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="5.0.0" />
+    <PackageReference Include="Microsoft.Extensions.Hosting.Abstractions" Version="5.0.0" />
+    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="5.0.0" />
+    <PackageReference Include="Microsoft.Extensions.ObjectPool" Version="5.0.0" />
```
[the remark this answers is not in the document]
evolvedlight: Not doing so seems to be more dangerous, having a higher version in the EF Core related libs, no? If a method is removed in v5 and the core library calls a method that is removed in v5, this would then fail?

src/providers/WorkflowCore.Providers.Elasticsearch/WorkflowCore.Providers.Elasticsearch.csproj
evolvedlight: Same answer as the others - we can try it but it seems worse to have different versions of the library in use

src/WorkflowCore/WorkflowCore.csproj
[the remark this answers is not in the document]
danielgerlag: No, if the consumer of `Workflow-Core` does not use the EF packages, the minimum version of the Abstraction packages could be 2.2.0, but if they add the EF packages that require 5.0, they'd be forced to update their own app to use 5.0, which should still be compatible with 2.2.0 of the abstractions that WorkflowCore depends on.

src/providers/WorkflowCore.Providers.Redis/WorkflowCore.Providers.Redis.csproj
danielgerlag: if the consumer of the redis package does not use the EF packages, the minimum version of the Abstraction packages could be 2.2.0, but if they add the EF packages that require 5.0, they'd be forced to update their own app to use 5.0, which should still be compatible with 2.2.0 of the abstractions that the redis package depends on.

evolvedlight:
@danielgerlag thanks, I've reverted those ones back down to their previous version of the abstractions library, and also bumped to the patch release 5.0.1 from 5.0.0

danielgerlag (approved):

PR #724, opened 2020-12-10
Added information on custom error handlers.

dimademi:
I have added some information regarding custom error handling.
I found myself digging through the code to understand how to do it, thought it would be useful in the docs.

### DIFF
--- a/docs/error-handling.md
+++ b/docs/error-handling.md
@@ -36,3 +36,9 @@ RetryInterval: '00:10:00'
 ## Global Error handling
 
 The WorkflowHost service also has a `.OnStepError` event which can be used to intercept exceptions from workflow steps on a more global level.
+
+## Custom Error Handlers 
+All `IWorkflowErrorHandler` are registered when `.AddWorkflow()` is called. 
+To create a custom error handler implement `IWorkflowErrorHandler` and add it as `Transient`.
+
+Each `IWorkflowErrorHandler` should specify `WorkflowErrorHandling` which suggests what type of errors it should handle.


dimademi:
One thing i wanted to add regarding this topic but couldn't find the way to do so -- 
How can you override the default exception handling for failed steps? 
e.g. If i wanted failed steps to terminate the workflow, and not retry

If any of you could let tell me the way to do it ill add it to the PR.

Cheers.

docs/error-handling.md
```diff
@@ -35,4 +35,10 @@ RetryInterval: '00:10:00'
 
 ## Global Error handling
 
-The WorkflowHost service also has a `.OnStepError` event which can be used to intercept exceptions from workflow steps on a more global level.
\ No newline at end of file
+The WorkflowHost service also has a `.OnStepError` event which can be used to intercept exceptions from workflow steps on a more global level.
+
+## Custom Error Handlers
```
danielgerlag, line 40: This is really an internal mechanism for handling error types, I suppose you could modify or extend it by registering other handlers in the IoC container, as you can do with any internal workflow-core service... however, I don't think this description really captures that or gives good examples on how to do it.  Maybe a better approach would be to write a doc on how to replace internal workflow-core services via IoC registration?  We could also add a reference to the new middleware feature on the error handling page?

dimademi: That's a good idea, I can open a new doc page for internal services that will, over time, contain all services documentation.
As for the middleware feature documentation - we can paste [https://github.com/danielgerlag/workflow-core/releases/tag/3.3](the detailed explanation you wrote here) into the docs

PR #727, opened 2020-12-13
Replace in memory GrayList with locks

glucaci:


danielgerlag (commented):
Hey, this looks really cool!  
I'm just not sure about the `QueueCache` name... do you have any other suggestions on what we could call it?

glucaci:
Hey, do you have a suggestion? I see it as a temporary cache for the queued items.

glucaci:
Hi @danielgerlag,

I did some work here but I analyzed the entire process and wanted to ask why we need in the end this caching and the `RunnablePoller`?

Because:
1. The workflow is pushed to the queue after is persisted (`WorkflowController.StartWorkflow`)
```csharp
...
string id = await _persistenceStore.CreateNewWorkflow(wf);
await _queueProvider.QueueWork(id, QueueType.Workflow);
...
```
2. The workflow is pushed again to the queue after is processed and in Running state (`WorkflowConsumer.ProcessItem`)
```csharp
if (workflow.Status == WorkflowStatus.Runnable)
{
    ...
    await _persistenceStore.PersistWorkflow(workflow);
    await QueueProvider.QueueWork(itemId, QueueType.Index);
    ...
}
```
I have found only one case in which the `RunnablePoller` is needed and this is when `NextExecution` is in the future, but this can be also covered in `WorkflowConsumer` in the way that we push the workflow again in the queue,

So the question is for which other case we are still using the `RunnablePoller`?

Thanks!

glucaci:
Hi @danielgerlag do you have any thoughts on this ?

danielgerlag:
sorry, I've been pretty busy... I will try spend some time drilling into this in the next few days.

glucaci:
Hi @danielgerlag, do you have any thoughts on this?

danielgerlag:
The polling was originally meant to be for future scheduling and a catch all, in case something was not queued or dropped from the queue because of locking, etc...
What was your idea for covering the future scheduled scenario in `WorkflowConsumer`?
For some providers, like redis, we can build a "unique" distributed queue to solve this, but it wouldn't work for all the providers.

glucaci:
Hi @danielgerlag, I've just simplify this one by replacing it with [IDistributedCache](https://docs.microsoft.com/en-us/aspnet/core/performance/caching/distributed?view=aspnetcore-5.0#idistributedcache-interface).

There are out there already some [implementations](https://docs.microsoft.com/en-us/aspnet/core/performance/caching/distributed?view=aspnetcore-5.0#establish-distributed-caching-services) from Microsoft and if needed anyone can implement `IDistributedCache` and register it.

By default it's already registered the InMemory implementation.

e.g. for Redis
```csharp
service
    .AddStackExchangeRedisCache(options => options.Configuration = connectionString)
    .AddWorkflow();
```

danielgerlag:
I was thinking, we could solve this problem by getting rid of the grey-list and using the distributed locks... when we enqueue a new workflow, we acquire a lock with a key like `queue:workflow:xxxx-xxxx-xxxx`, and when we dequeue it, we release the lock.... this would ensure the same item isn't in the queue multiple times.  I also means nobody has to add additional infra their environments to resolve this.
The one issue to think about is, if the queue is purged or something, we also need those locks to be released... maybe if they had a TTL or something?

src/WorkflowCore/Services/BackgroundTasks/RunnablePoller.cs
```diff
@@ -51,27 +49,14 @@ private async void PollRunnables(object target)
         {
             try
             {
-                if (await _lockProvider.AcquireLock("poll runnables", new CancellationToken()))
+                _logger.LogInformation("Polling for runnable workflows");
```
glucaci, line 52: We don't need here to create a generic lock for the hole polling process. Each single item will be protected by the "wf:item" lock. So multiple processes can safely enqueue concurrently items in the queue

glucaci:
@danielgerlag I think also that is a much nicer solution and I did the changes.

I think the complexity of implementing the TTL for the locks doesn't payoff any we could document it, that if someone purge the items queue should also purge the locks.

glucaci:
If #851 gets merged, I have to update here the usage also.

src/WorkflowCore/Services/BackgroundTasks/RunnablePoller.cs
```diff
@@ -51,27 +49,14 @@ private async void PollRunnables(object target)
...
+                var runnables = await _persistenceStore.GetRunnableInstances(_dateTimeProvider.Now);
+                foreach (var item in runnables)
                 {
-                    try
+                    if (await _lockProvider.AcquireLock($"wf:{item}", default))
```
danielgerlag, line 56: Do you think we should use a more specific lock key like `queue:wf:xxxx`?
ie. have different locks for queueing and processing?

danielgerlag (approved):

danielgerlag:
@glucaci  One side-effect to consider is that some of the distributed lock implementations send a heartbeat to keep the lock alive... if the number of enqueued items grows large, that could be a hefty operation to keep sending heartbeats for all those locks?

danielgerlag (commented):
I spent some more time thinking about this one...
I don't think this would work because many of the lock providers also track the lock locally on the node that originally requested it and auto-renew it... so unless THAT same node releases it, it will keep getting auto-renewed.  In addition, some of the lock providers can only release locks from the same node that issued them (Azure Blobs, AWS, SQL, etc.)
I think we'd need to push the responsibility to each queueing implementation to ensure we have a "unique queue", eg. the Azure storage queue could also use blob storage to ensure uniqueness, etc.

PR #732, opened 2020-12-17
fix adding a key to dictionary in thread unsafe manner

georgebratu:
Hi ! Once in a while, we were getting an error "the same key was already added" in the QueueConsumer.cs It seems that the adding a key in the dictionary was not written to be thread safe. So this pull request fixes this concurrency issue. So the only change is that before trying to add "item" in key in "_activeTasks" check inside the lock if it's already added. If it's not the carry on, otherwise, go back to the initial check.

### DIFF
--- a/src/WorkflowCore/Services/BackgroundTasks/QueueConsumer.cs
+++ b/src/WorkflowCore/Services/BackgroundTasks/QueueConsumer.cs
@@ -83,27 +83,45 @@ namespace WorkflowCore.Services.BackgroundTasks
                         continue;
                     }
 
-                    var hasTask = false;
-                    lock (_activeTasks)
-                    {
-                        hasTask = _activeTasks.ContainsKey(item);
-                    }
-                    if (hasTask)
-                    {
-                        _secondPasses.Add(item);
-                        if (!EnableSecondPasses)
-                            await QueueProvider.QueueWork(item, Queue);
-                        continue;
-                    }                   
-
-                    _secondPasses.TryRemove(item);
-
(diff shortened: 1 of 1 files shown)


georgebratu:
Any thoughts on this ? Do you need more information ?

danielgerlag (commented):
What about just using `ConcurrentDictionary` from `System.Collections.Concurrent`?

danielgerlag:
I believe this issue was fixed in a previous release

georgebratu:
Sorry about late reply, using ConcurrentDictionary didnt fix it. It's clear that the code is not thead safe if you dont implement the double check properly.
You are doing:

```
lock (_activeTasks)
                    {
                        hasTask = _activeTasks.ContainsKey(item);
                    }

......... some code here .....

lock (_activeTasks)
                    {
                        _activeTasks.Add(item, waitHandle);
                    }
```

So when you add the item in activeTasks it was already added in another thread.
That's why I inserted a check before adding the key like:

```
checkAgainTheDictionary:
lock (_activeTasks)
                    {
                        hasTask = _activeTasks.ContainsKey(item);
                    }

......... some code here .....

lock (_activeTasks)
                    {
                        if (_activeTasks.ContainsKey(item))
                            goto checkAgainTheDictionary;
                        _activeTasks.Add(item, waitHandle);
                    }
```

You ALWAYS need to check inside the lock block if the key already exists before adding in order to do proper thread safe code.

danielgerlag:
https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/lock-statement

PR #749, opened 2021-01-12
fix of using session registered event handler

jiripoledna:
I try to fix the problem with session in service bus, I think there is a mistakes,

- the session id has to be there
- the complete or abandon has to be called using session object not the subscribtion


danielgerlag (changes requested):
Instead of setting the session ID when sending, why don't we use `RegisterMessageHandler` instead of `RegisterSessionHandler` when we start listening for messages?

jiripoledna:
The code leads me to understad that it needs session, there is used session options, session handler and so on. I can rewrite it to use message handler and then it will probably work, but there is more code changes then this few lines, The code which is now in actual version is mix of session and non session aproach and it will not work at all. 

if you decide to use session approach only question is what is correct sessionId. I put there something.
But anyway the library is super I'm testing it over SQL persistence, but service bus is more suitable to me

PR #757, opened 2021-01-21
Fix testing with DI in Workflow

xlegalles:
Dependency injection is supported in Workflow but cannot be tested (fix)
It is related to changes made in PR: Allow dependency injection on Workflow #576
The new() constraint on TWokflow is still present on WorkflowTest while it has been removed elsewhere

### DIFF
--- a/test/WorkflowCore.IntegrationTests/Scenarios/DiScenario.cs
+++ b/test/WorkflowCore.IntegrationTests/Scenarios/DiScenario.cs
@@ -87,6 +87,8 @@ namespace WorkflowCore.IntegrationTests.Scenarios
     {
         protected void ConfigureHost(IServiceProvider serviceProvider)
         {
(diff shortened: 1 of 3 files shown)


xlegalles:
It may also be useful to add comments in the documentation: 
- "step and workflow classes must be registered in the DI container if they expect dependencies to be injected". This is obvious, but personally I was expecting that it was already done by default.
- "step classes must inherit directly from StepBody/StepBodyAsync i.e. cannot inherit from intermediate classes". No factorization allowed as I've started to do.

Thank you

daniele-betsson (approved):

daniele-betsson:
would be fantastic to have this merged some years later :D

wfcore-manager:
Closing as stale (opened 2021, long inactive). Please reopen if still relevant.

PR #770, opened 2021-02-06
Error middleware enhancement

MrZoidberg:
This PR is adding a WorkflowInstance parameter to `IWorkflowMiddlewareErrorHandler` interface so that it's possible to log only exception, but also related workflow information like ID:

```
_log.LogError(ex, "An error occurred running workflow '{workflow}' middleware: {Message}", workflowInstance.Id, ex.Message);
```

WARNING: this is a breaking change in terms that any 3rd party implementations of IWorkflowMiddlewareErrorHandler won't work without changes.

### DIFF
--- a/src/WorkflowCore/Interface/IWorkflowMiddlewareErrorHandler.cs
+++ b/src/WorkflowCore/Interface/IWorkflowMiddlewareErrorHandler.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Threading.Tasks;
+using WorkflowCore.Models;
 
 namespace WorkflowCore.Interface
 {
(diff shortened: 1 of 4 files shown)


src/WorkflowCore/Interface/IWorkflowMiddlewareErrorHandler.cs
```diff
@@ -11,8 +12,9 @@ public interface IWorkflowMiddlewareErrorHandler
         /// <summary>
         /// Asynchronously handle the given exception.
         /// </summary>
+        /// <param name="workflowInstance">Workflow instance where error happened</param>
         /// <param name="ex">The exception to handle</param>
         /// <returns>A task that completes when handling is done.</returns>
-        Task HandleAsync(Exception ex);
+        Task HandleAsync(WorkflowInstance workflowInstance, Exception ex);
```
danielgerlag, line 18: How can we make this backward compatible?

danielgerlag (changes requested):

src/WorkflowCore/Interface/IWorkflowMiddlewareErrorHandler.cs
MrZoidberg: @danielgerlag I can introduce something like `IWorkflowMiddlewareErrorHandlerV2` that will have an old and a new method. And update the usages. However, I don't like this approach. What do you think?

PR #775, opened 2021-02-14
Add execute middleware phase

glucaci:
Hi,

I want to be able to execute post middleware in any workflow status case and not only when is completed.
What do you think? Would this be enough?

We want to build metrics using the post middleware. (execution time, who many workflow completed, suspended, terminated, etc.)

Thanks!

### DIFF
--- a/src/WorkflowCore/Services/WorkflowExecutor.cs
+++ b/src/WorkflowCore/Services/WorkflowExecutor.cs
@@ -96,12 +96,12 @@ namespace WorkflowCore.Services
                 _cancellationProcessor.ProcessCancellations(workflow, def, wfResult);
             }
             ProcessAfterExecutionIteration(workflow, def, wfResult);
-            DetermineNextExecutionTime(workflow, def);
+            await DetermineNextExecutionTime(workflow, def);
 
             using (var scope = _serviceProvider.CreateScope())
             {
                 var middlewareRunner = scope.ServiceProvider.GetRequiredService<IWorkflowMiddlewareRunner>();
-                await middlewareRunner.RunPostMiddleware(workflow, def);
+                await middlewareRunner.RunExecuteMiddleware(workflow, def);
             }
 
             return wfResult;
@@ -212,7 +212,7 @@ namespace WorkflowCore.Services
             }
         }
 
-        private void DetermineNextExecutionTime(WorkflowInstance workflow, WorkflowDefinition def)
+        private async Task DetermineNextExecutionTime(WorkflowInstance workflow, WorkflowDefinition def)
         {
             //TODO: move to own class
             workflow.NextExecution = null;
@@ -256,6 +256,12 @@ namespace WorkflowCore.Services
 
             workflow.Status = WorkflowStatus.Complete;
             workflow.CompleteTime = _datetimeProvider.UtcNow;
+
+            using (var scope = _serviceProvider.CreateScope())
+            {
+                var middlewareRunner = scope.ServiceProvider.GetRequiredService<IWorkflowMiddlewareRunner>();
+                await middlewareRunner.RunPostMiddleware(workflow, def);
+            }
 
             _publisher.PublishNotification(new WorkflowCompleted
             {
--- a/src/WorkflowCore/Services/WorkflowExecutor.cs
+++ b/src/WorkflowCore/Services/WorkflowExecutor.cs
@@ -212,7 +212,10 @@ namespace WorkflowCore.Services
             workflow.NextExecution = null;
 
             if (workflow.Status == WorkflowStatus.Complete)
+            {
+                await RunPostMiddleware(workflow, def);
                 return;
+            }
 
             foreach (var pointer in workflow.ExecutionPointers.Where(x => x.Active && (x.Children ?? new List<string>()).Count == 0))
             {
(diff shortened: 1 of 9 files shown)


danielgerlag:
@dflor003

glucaci:
any news ? @danielgerlag @dflor003

dflor003:
Hey! Apologies for not looking at this sooner. So with this change, have you verified that it will report for canceled and terminated workflows? If so, sounds like it shouldn't break anything.

glucaci:
I think so, maybe @danielgerlag kann confirm that this method reach the code in any case

dflor003:
Would it make sense to add an integration test to confirm that this is the case?

dflor003:
@danielgerlag This PR good to merge? I'm good with it if you are. My team is also in need of this fix.

glucaci:
We could follow-up with a test.
I didn't had time to write one, but we could easily extend some integration tests to cover also this.

glucaci:
@danielgerlag can we merge this PR ?

src/WorkflowCore/Services/WorkflowExecutor.cs
```diff
@@ -242,18 +250,14 @@ private async Task DetermineNextExecutionTime(WorkflowInstance workflow, Workflo
...
 
             if ((workflow.NextExecution != null) || (workflow.ExecutionPointers.Any(x => x.EndTime == null)))
+            {
                 return;
+            }
 
             workflow.Status = WorkflowStatus.Complete;
             workflow.CompleteTime = _datetimeProvider.UtcNow;
 
-            using (var scope = _serviceProvider.CreateScope())
```
danielgerlag, line 250: Will moving this break existing expectations on when this fires? ie. this could break existing use cases out there?

glucaci:
@danielgerlag just fixed the conflicts, can we merge this?

src/WorkflowCore/Services/WorkflowExecutor.cs
danielgerlag: @glucaci ^

danielgerlag:
I think, moving when this fires could break existing use cases out there. This is not backward compatible.
I think this needs to be a new middleware "type"

danielgerlag (changes requested):
I think, moving when this fires could break existing use cases out there. This is not backward compatible.
I think this needs to be a new middleware "type"

src/WorkflowCore/Services/WorkflowExecutor.cs
glucaci: Somehow wrote a comment yesterday and is not here anymore...
My opinion is that this is not a breaking change because everything that is public to the client is the `workflow` itself and the `definition`, so is nothing that we changed in the public API.

Also moving the position where the `scope` gets created is not a breaking change.

Can you detailed what did you think that is creating a breaking change?

glucaci:
I don't think that is a breaking change, see my comment on code

src/WorkflowCore/Services/WorkflowExecutor.cs
danielgerlag: What about consumers of workflow-core that are using the post middleware and only expecting it to run on workflow completion. Now it is going to run every time a step executes, multiple times throughout the workflow's life.

glucaci: Yes, makes sens, I read again the docu: ` get run after the workflow has completed`.
In this case we can create another one and call it like above.

Do you have any naming preference?
Proposal: `WorkflowMiddlewarePhase.OnChange`

glucaci:
@danielgerlag can you take a look?

danielgerlag (commented):
This is great, it preserves the existing behavior.

I am curious though, this implementation doesn't match what you described in the original submission, which sounded like the PostWorkflow middleware should also run when workflows change state, is this intentional?

Another question have, is could your use case be achieved with step middleware?

Also, could we one or two scenario tests to ensure the middleware is firing the expected number of times throughout a workflow's lifecycle?

glucaci:
> I am curious though, this implementation doesn't match what you described in the original submission, which sounded like the PostWorkflow middleware should also run when workflows change state, is this intentional?

I think is not the same what i've wrote because now is running on each workflow execution, also when the workflow remains in "Running" state because has another step with next execution point in future. 
But it's much better because what we want to achieve with it is that we build some insights without pooling the data from database.  On each run we will push different events on a bus based on workflow state and than can build insights like, "who many times an workflow was executing", "how many completed or terminated workflow we have", etc.

> Another question have, is could your use case be achieved with step middleware?

I don't think because in a step middleware you don't have the final workflow state from that run, right? I need to be sure that when i get called into this middleware the workflow will not be changed anymore, actually i need the state of the workflow which will be save in database.

> Also, could we one or two scenario tests to ensure the middleware is firing the expected number of times throughout a workflow's lifecycle?

I wrote [here](https://github.com/danielgerlag/workflow-core/pull/775/files#diff-07374253949095aa15bb1602756076b9c16128b3e9f6a2e3c21dbc5e51227075R119) one and i can write one or two more when the workflow is changing the state or when is not changing the state, but I don't know if is not to much because you can build a lot of scenarios.

danielgerlag (approved):
Sounds good, increment the minor version number and we can get this published.

glucaci:
Hi @danielgerlag, I've incremented the version to `3.4.0` and added release notes with a sample.

danielgerlag:
Nice work @glucaci

PR #776, opened 2021-02-14
Fix workflow status MongoDB serialization

glucaci:
Because we use a explicit property map MongoDB will not take in consideration the `EnumRepresentationConvention`, that's why we have to set explicit the serializer.

### DIFF
--- a/src/providers/WorkflowCore.Persistence.MongoDB/Services/MongoPersistenceProvider.cs
+++ b/src/providers/WorkflowCore.Persistence.MongoDB/Services/MongoPersistenceProvider.cs
@@ -7,6 +7,7 @@ using System.Linq;
 using System.Threading.Tasks;
 using MongoDB.Bson;
 using MongoDB.Bson.Serialization.Conventions;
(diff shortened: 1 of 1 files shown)


src/providers/WorkflowCore.Persistence.MongoDB/Services/MongoPersistenceProvider.cs
```diff
@@ -44,7 +45,8 @@ static MongoPersistenceProvider()
                 x.MapProperty(y => y.WorkflowDefinitionId);
                 x.MapProperty(y => y.Version);
                 x.MapProperty(y => y.NextExecution);
-                x.MapProperty(y => y.Status);
+                x.MapProperty(y => y.Status)
+                    .SetSerializer(new EnumSerializer<WorkflowStatus>(BsonType.String));
```
danielgerlag, line 49: is this backward compatible with previously persisted workflows?

glucaci: yes, it is

Issue #773, opened 2021-02-11
Event is not processed, but marked as processed


MrZoidberg:
I have a serious problem with events.
I'm using a multi-node environment with MongoDb and Redis (locking, queues)

Here what I have currently investigated:

1. There is a workflow which waits for the event to be triggered. Currently, the last executing pointer is waiting for event (reduced to keep IP safe):
```json
{
        "_id": "c16d5b41-481a-4dde-9730-1cf7c5f5607b",
        "StepId": 418,
        "Active": false,
        "SleepUntil": null,
        "PersistenceData": null,
        "StartTime": {
            "$date": "2021-02-11T02:06:11.055Z"
        },
        "EndTime": null,
        "EventName": "EVENT_NAME",
        "EventKey": "P-hTcAdFSTQ-d8f23ce3-c1cf-4f53-b736-f6bc29493e2e",
        "EventPublished": false,
        "EventData": null,
        "ExtensionAttributes": {},
        "StepName": null,
        "RetryCount": 0,
        "Children": [],
        "ContextItem": null,
        "PredecessorId": "a6fb840e-3a2a-4cdb-a23b-645a962d7861",
        "Outcome": null,
        "Status": "WaitingForEvent",
        "Scope": ["c39eda9f-b016-4b8b-a2d5-611ec9088206", "0eb4ca0d-ec92-46df-8d7c-14e39918e105", "61debfc2-70a7-415b-9c38-99c9af286f36", "a858fd08-8a4c-4141-bf5b-87018c7b9dbc", "4fae292b-beee-479e-bd1f-36df291f8b3d", "ea5609e1-1831-4a75-8315-7442716a0c3f", "f01ae33c-3b3f-438a-877d-8a4e8e5b6ca4", "bd2f6dc5-17ba-4861-aa52-1272d1c1f43d", "ce5c51aa-d0e3-4692-8d09-4cfb89539e0f", "347f9f6f-a87c-422c-a25d-30570809528d"]
    }
```
2. There is an active subscription for the event, so it looks like it's still waiting for the event:
```json
{
   "_id":"602491134d7b14f3227b9e50",
   "EventName":"EVENT_NAME",
   "EventKey":"P-hTcAdFSTQ-d8f23ce3-c1cf-4f53-b736-f6bc29493e2e",
   "StepId":418,
   "ExecutionPointerId":"c16d5b41-481a-4dde-9730-1cf7c5f5607b",
[...]

PR #782, opened 2021-02-23
Don't mark an event as processed when there are no subscriptions

MrZoidberg:
Mostly fixes issue #773

### DIFF
--- a/src/WorkflowCore/Services/BackgroundTasks/EventConsumer.cs
+++ b/src/WorkflowCore/Services/BackgroundTasks/EventConsumer.cs
@@ -68,16 +68,20 @@ namespace WorkflowCore.Services.BackgroundTasks
                     }
 
                     var toQueue = new HashSet<string>();
-                    var complete = true;
 
-                    foreach (var sub in subs.ToList())
-                        complete = complete && await SeedSubscription(evt, sub, toQueue, cancellationToken);
+                    if (subs.Any())
+                    {
+                        var complete = true;
 
-                    if (complete)
-                        await _eventRepository.MarkEventProcessed(itemId);
+                        foreach (var sub in subs.ToList())
+                            complete = complete && await SeedSubscription(evt, sub, toQueue, cancellationToken);
 
-                    foreach (var eventId in toQueue)
-                        await QueueProvider.QueueWork(eventId, QueueType.Event);
+                        if (complete)
+                            await _eventRepository.MarkEventProcessed(itemId);
+
+                        foreach (var eventId in toQueue)
+                            await QueueProvider.QueueWork(eventId, QueueType.Event);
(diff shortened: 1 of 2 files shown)


danielgerlag:
What about setting the effective datetime of the `WaitFor` to an earlier time?

MrZoidberg:
@danielgerlag I'm not using time in WaitFor, so it is set to minimal time

danielgerlag:
In your example you have
```c#
public void Build(IWorkflowBuilder<MyDataClass> builder)
            {
                builder
                    .StartWith(context => ExecutionResult.Next())
                    .WaitFor("MyEvent", data => data.Value1.ToString(), data => DateTime.Now)
                    .Output(data => data.Value2, step => step.EventData)
                    .Then(context => ExecutionResult.Next())
                    .Output(data => data.Value3, step => "OK");
            }
```
What if you used

.WaitFor("MyEvent", data => data.Value1.ToString(), data => **DateTime.MinValue**)

MrZoidberg:
@danielgerlag I see. I won't help because the EventConsumer is executed when there are no subscriptions yet, so the time there doesn't matter.

danielgerlag:
@MrZoidberg we marked older events as unprocessed again when creating back dated subscriptions
see: https://github.com/danielgerlag/workflow-core/blob/master/src/WorkflowCore/Services/BackgroundTasks/WorkflowConsumer.cs#L95

MrZoidberg:
@danielgerlag please check the test, it doesn't work with `.WaitFor("MyEvent", data => data.Value1.ToString(), data => DateTime.MinValue)`. Maybe I'm missing something, but in my real app I always use `DateTime.MinValue` and with the real scenario sending an event to Kafka and receiving a response very quickly the event is marked as processed and is never marked as unprocessed again. btw in the Workflow consumer, there is no lock for the event, so maybe the problem is that event is marked as processed after it is marked as unprocessed, what do you think?

unaizorrilla:
We are experimenting the same kind of issue!

MrZoidberg:
@danielgerlag hey. just checking if we can continue the discussion on this topic. I'm using a fork due to this issue and would love to help you to fix it.

VKAlwaysWin:
@danielgerlag one note here that there are no locking on event by event Id here - https://github.com/danielgerlag/workflow-core/blob/master/src/WorkflowCore/Services/BackgroundTasks/WorkflowConsumer.cs#L94 which may lead to unexpected behaviour. Maybe it is worth to add something like this: https://github.com/danielgerlag/workflow-core/blob/master/src/WorkflowCore/Services/BackgroundTasks/EventConsumer.cs#L36 .

MrZoidberg:
Closing my PR in flavor of #851

PR #786, opened 2021-03-02
fix event greylist stuck

### LINKED ISSUE #722: Workflow Stuck Waiting for Greylisted Event
Hi, we have an interesting scenario where we need to coordinate events across multiple workflows and we are having some issues where events are getting greylisted and sometimes never get processed.

First, let me describe our setup:

We have 2 different types of workflows, lets say the first one is called `CoordinatorWorkflow` and the second is called `SubTaskWorkflow`. For a given task, there will be exactly 1 `CoordinatorWorkflow` spun up and N `SubTaskWorkflow`s. 

When `CoordinatorWorkflow` is spun up, it knows how many `SubTaskWorkflow`s there are that correspond to it and has a unique identifier for each of the `SubTaskWorkflow` in the "set".

As an example, let's say we have a set of 2 `SubTaskWorkflow`s identified by `SubtaskA` and `SubtaskB` that both start around the same time and `CoordinatorWorkflow` is passed a set of `["SubTaskA", "SubTaskB"]`.

The very first thing that `CoordinatorWorkflow` does is go into a for loop on `["SubTaskA", "SubTaskB"]` and waits for an event of type `SubTaskFinished` with the key being each of the identifiers `SubTaskA` and `SubTaskB`.

Eventually each of the `SubTaskWorkflow`s gets to a certain point and publishes an event of type `SubTaskFinished` with its identifier (`SubTaskA` or `SubTaskB`) as the key. At this point they wait for an event of type `CoordinationFinished`.

Once `CoordinatorWorkflow` received the events from each of the `SubTaskWorkflow`, it then proceeds to do some work and then fires off a `CoordinationFinished` event and completes.

Each `SubTaskWorkflow` then gets the `CoordinationFinished` event and then proceeds to do some more work and completes.

This is roughly the flow we are trying to achieve and have gotten most of the way there (Note: These aren't the actual names of the workflows, but I've tried to make them somewhat generic and domain-agnostic to simplify it).

The problem we are getting, however, is that at some point in the process, we get stuck waiting for events that never arrive. The events do indeed get published but we see a bunch of messages in the logs like the following and the workflows waiting for the events never have a chance to process them.

```bash
[16:06:44 DBG] Got greylisted event evt:{Id}
```

Any ideas as to what we may be doing wrong? Is this a known issue? If so, any work arounds?

Here are a few other observations that my team saw while troubleshooting this:

- This happens sporadically (but fairly frequently) when running locally with 1 workflow node, but we do not see this in our development and production environments where we run 2 and 4 workflow engine nodes respectively. I have a feeling this is because `IGreyList` is registered as an in-memory singleton and the other workflow processes don't have it grey listed so they are free to pick up those events.
- We recently added some integration tests around the process outlined above and we also ran into this issue. Work around in the integration tests was to provide our own fake implementation of `IGreyList` that basically ignored any items prefixed with `evt`.
- I did some digging in the code and it looks like while there is something that inserts events into the grey list, there doesn't seem to be anything that removes events from the greylist.

anderr225:
fixes #722

### DIFF
--- a/src/WorkflowCore/Services/BackgroundTasks/EventConsumer.cs
+++ b/src/WorkflowCore/Services/BackgroundTasks/EventConsumer.cs
@@ -107,6 +107,7 @@ namespace WorkflowCore.Services.BackgroundTasks
             if (!await _lockProvider.AcquireLock(sub.WorkflowId, cancellationToken))
             {
                 Logger.LogInformation("Workflow locked {0}", sub.WorkflowId);
+                await QueueProvider.QueueWork(evt.Id, QueueType.Event);
                 return false;
             }
             
--- a/src/WorkflowCore/Services/BackgroundTasks/EventConsumer.cs
+++ b/src/WorkflowCore/Services/BackgroundTasks/EventConsumer.cs
@@ -36,6 +36,7 @@ namespace WorkflowCore.Services.BackgroundTasks
             if (!await _lockProvider.AcquireLock($"evt:{itemId}", cancellationToken))
             {
                 Logger.LogInformation($"Event locked {itemId}");
+                await QueueProvider.QueueWork(itemId, QueueType.Event);
                 return;
             }
             
--- a/src/WorkflowCore/Services/BackgroundTasks/RunnablePoller.cs
+++ b/src/WorkflowCore/Services/BackgroundTasks/RunnablePoller.cs
@@ -93,7 +93,6 @@ namespace WorkflowCore.Services.BackgroundTasks
                             if (_greylist.Contains($"evt:{item}"))
                             {
                                 _logger.LogDebug($"Got greylisted event {item}");
-                                _greylist.Add($"evt:{item}");
                                 continue;
                             }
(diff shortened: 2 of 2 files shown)


src/WorkflowCore/Services/BackgroundTasks/EventConsumer.cs
```diff
@@ -36,6 +36,7 @@ protected override async Task ProcessItem(string itemId, CancellationToken cance
             if (!await _lockProvider.AcquireLock($"evt:{itemId}", cancellationToken))
             {
                 Logger.LogInformation($"Event locked {itemId}");
+                await QueueProvider.QueueWork(itemId, QueueType.Event);
```
danielgerlag, line 39: Could this cause a [poison message](https://en.wikipedia.org/wiki/Poison_message)?

anderr225: This line may be. Actually it looks like adding to a queue when failing to acqure an evt lock is a bad idea. But it is necessary when failing to acquire a workflow lock.

PR #787, opened 2021-03-08
add ProceedOnCancel to DSL and add workflowStepParameter to CancelCondition

Nokecy:


Nokecy:
@danielgerlag  @dflor003

PR #810, opened 2021-04-14
Add components overview

Granjow:
This documentation PR adds a short overview over some main components which are available in Workflow Core. Having a bigger picture or architecture view was something I missed when starting with Workflow Core.

I'd prefer something like a C4 components diagram, but readthedocs seems to be limited with diagram options.

Any suggestions/ideas?


docs/getting-started.md
```diff
@@ -114,6 +114,16 @@ Console.ReadLine();
...
 ```
 
+## Registry
+
+The workflow host keeps workflow definitions in the registry when they are registered.  When starting a workflow, the workflow host creates a new instance for the desired workflow definition.
+
+
+## Persistence
+
+The persistence provider persists the state of a workflow instance with execution pointers.  They are created when visiting a step while executing a workflow and hold information about their outcome.
```
danielgerlag, line 124: I think execution pointers are more of an internal concept, not really specific to persistence?
Either way, I'm not sure this paragraph gives a description of what they are.

docs/getting-started.md
```diff
@@ -114,6 +114,16 @@ Console.ReadLine();
 host.Stop();
 ```
 
+## Registry
+
+The workflow host keeps workflow definitions in the registry when they are registered.  When starting a workflow, the workflow host creates a new instance for the desired workflow definition.
```
danielgerlag, line 119: The workflow host does not keep the definitions, the registry does, the workflow host accesses the registry.

danielgerlag (commented):
Is the target audience consumers of workflow core or contributors to the library itself?

Granjow:
@danielgerlag Thanks for your review, I have updated the pull request. Better now?

Target audience is mainly consumers of the library, but I have found it easier to work with Workflow Core when also understanding some of the more internal concepts.

wfcore-manager:
Closing as abandoned (draft, no updates since 2021). Please reopen or open a new PR if you still want this.

PR #811, opened 2021-04-15
fix cosmos client leak

brucezlata:
@danielgerlag I have a manual test from local using webapi sample. And monitor socket count using netstat.

I do see there is missing for test for azure provider. would be a next?

Could you please kindly help review this PR to fix cosmos client memory leak?

# Add CosmosDbClient which implements a interface for DI
# Use singleton for CosmosDbClient and it will be disposed by dotnet  container
# Fix EnsureStoreExists  deaklock risk

**Describe the change**
A clear and concise description of what the change is. Any PR submitted without a description of the change will not be reviewed.

**Tests**

**Breaking change**

**Additional context**
Any additional information you'd like to provide?


src/providers/WorkflowCore.Providers.Azure/Services/CosmosDbClient.cs
```diff
@@ -0,0 +1,48 @@
...
+using Microsoft.Azure.Cosmos;
+using WorkflowCore.Providers.Azure.Interface;
+
+namespace WorkflowCore.Providers.Azure.Services
+{
+    public class CosmosDbClient : ICosmosDbClient, IDisposable
+    {
+        private bool isDisposed = false;
+
+        public CosmosClient _client;
```
VKAlwaysWin, line 11: Should it be private ?

brucezlata: got. thanks.

src/providers/WorkflowCore.Providers.Azure/Interface/ICosmosDbClient.cs
```diff
@@ -0,0 +1,9 @@
+﻿using Microsoft.Azure.Cosmos;
+
+namespace WorkflowCore.Providers.Azure.Interface
+{
+    public interface ICosmosDbClient
```
danielgerlag, line 5: Could we name this `ICosmosClientFactory`?

src/providers/WorkflowCore.Providers.Azure/ServiceCollectionExtensions.cs
```diff
@@ -28,8 +28,9 @@ public static WorkflowOptions UseAzureSynchronization(this WorkflowOptions optio
 
         public static WorkflowOptions UseCosmosDbPersistence(this WorkflowOptions options, string connectionString, string databaseId)
         {
-            options.Services.AddTransient<ICosmosDbProvisioner>(sp => new CosmosDbProvisioner(connectionString, sp.GetService<ILoggerFactory>()));
-            options.UsePersistence(sp => new CosmosDbPersistenceProvider(connectionString, databaseId, sp.GetService<ICosmosDbProvisioner>()));
+            options.Services.AddSingleton<ICosmosDbClient>(sp => new CosmosDbClient(connectionString));
+            options.Services.AddTransient<ICosmosDbProvisioner>(sp => new CosmosDbProvisioner(sp.GetService<ICosmosDbClient>(), sp.GetService<ILoggerFactory>()));
```
danielgerlag, line 32: Do we still need the lambda here?

src/providers/WorkflowCore.Providers.Azure/Services/CosmosDbPersistenceProvider.cs
```diff
@@ -19,19 +19,19 @@ public class CosmosDbPersistenceProvider : IPersistenceProvider
 
         private ICosmosDbProvisioner _provisioner;
         private string _dbId;
-        private CosmosClient _client;
+        private ICosmosDbClient _client;
```
danielgerlag, line 22: Should this be named to indicate it's a client factory rather than a client itself?

brucezlata: Make sense. Thanks, a lot.

src/providers/WorkflowCore.Providers.Azure/Interface/ICosmosDbClient.cs
brucezlata: Great. Like ICosmosClientFactory

src/providers/WorkflowCore.Providers.Azure/ServiceCollectionExtensions.cs
brucezlata: I think yes. Let it disposal automatically.
See
https://docs.microsoft.com/en-us/dotnet/core/extensions/dependency-injection#service-registration-methods

danielgerlag: I mean
```c#
 options.Services.AddTransient<ICosmosDbProvisioner>();
```

brucezlata:
@danielgerlag @VKAlwaysWin Is that ok to approve this PR and merge to master?

VKAlwaysWin:
@brucezlata it makes sense for me. Thanks.
@danielgerlag may we promote this PR and merge it ? I think this may be important for users with CosmosDb as Persistant storage.

danielgerlag (approved):
Looks good. Can we increment the minor version number?, then we can publish.

brucezlata:
> Looks good. Can we increment the minor version number?, then we can publish.

Thanks, @danielgerlag. I increment the minor version number.

PR #813, opened 2021-04-19
Add transaction support

glucaci:
Update:
We have created a Transaction Scope extension package for MongoDB Driver and integrated in WorkflowCore
https://www.nuget.org/packages/MongoDB.Extensions.Transactions

_______
Hi @danielgerlag, we want to be able to use transactions when starting a workflow. There are multiple ways to do it, also with or without breaking change.

I opened this PR to start discussing which API fit's better and implement it at least for MongoDB. This draft API should not bring any breaking change. If this way fit's for you I will start implementing it completely.

In order to stay away from generics and introduce breaking changes, I've used the `object` type on the `Session`.

I created a small sample who this could look. (it's only sample, not real world :) )
```csharp
public class MongoTransaction : ITransaction
{
    private readonly IMongoClient _client;
    private IClientSessionHandle _session;

    public MongoTransaction(IMongoClient client)
    {
        _client = client;
    }

    public object Session => _session;

    public async Task StartSessionAsync()
    {
        _session = await _client.StartSessionAsync();
        _session.StartTransaction();
    }

    public async Task CommitAsync()
    {
        await _session.CommitTransactionAsync();
    }
}

public class MyService
{
    private readonly IWorkflowHost _workflowHost;
    private readonly IMongoClient _client;

    public MyService(IWorkflowHost workflowHost, IMongoClient client)
    {
        _workflowHost = workflowHost;
        _client = client;
    }

    public async Task ExecuteAsync()
    {
        var transaction = new MongoTransaction(_client);
        await transaction.StartSessionAsync();

        // do actions with transaction
        await _workflowHost.StartWorkflow(new MyData(), transaction: transaction);
        // do more actions with same transaction

        await transaction.CommitAsync();
    }
}
```

_Update_: I've converted it to draft until we are finishing the discussion about API.


glucaci:
@danielgerlag any thoughts on this one?

danielgerlag:
Looks good, but what do you think of using the built-in transaction scope?
https://docs.microsoft.com/en-us/dotnet/api/system.transactions.transactionscope

glucaci:
@danielgerlag MongoDB has no support for `TransactionScope` that's why I tought we could implement a flexible transaction support on the library level.
Although for example you can use `TransactionScope` with `EntityFramework` which supports it.

My proposal will extend the transaction support for the persistence providers which are not supporting `TransactionScope`.

What do you thing?

vladimir-kovalyuk:
What about managing transactions via ambient context? Especially taking into account that WFC work with SQL  through EFCore.
For instance the most thought out approach I found in @mehdime's https://github.com/mehdime/DbContextScope. The idea can be re-worked to accommodate the other persistence providers.

glucaci:
We are working on a extension to support `TransactionScope` for MongoDB. I will integrate it in ``WorkflowCore` once we have it done.

glucaci:
@danielgerlag do you have time to review this please?

danielgerlag:
@glucaci thanks... is `MongoDB.Extensions.Transactions` your library? Do you have some docs on how it's using transactions under the hood?

glucaci:
@danielgerlag yes, it's our open source library which is part of multiple MongoDB extensions (https://github.com/SwissLife-OSS/mongo-extensions)
@PascalSenn did the most work on it, it's pretty fresh, so we are still working on documentation.

PascalSenn:
@danielgerlag 
We wrap the IMongoCollection with a Proxy object. 
The proxy object checks if there is a open transaction and creates a session based on the transaction id. 
The proxy then just always uses the same session based on the transaction id.

glucaci:
Hi @danielgerlag i changed to a "non invasive" way by providing a new extension method for registering MongoDB and we will extend our package to can register the mongo transaction on the `MongoClient` or on the `IMongoDatabase` level.

Hope that's fine for you.

Thanks!

PR #830, opened 2021-05-14
Retry failed workflow

### LINKED ISSUE #829: How to retry a workflow, of which one step failed?
Occasionally, a step may fail because of network connectivity or any issue. I do not see anywhere in the docs a possibility to restart the workflow from the step it failed during its previous run. 

Resume workflow only resumes previously suspended workflows.

DevsAnon:
**Describe the change**
When workflow is terminated for a step failure. This functionality will allow the possibility to retry the workflow from the failed step. 

**Additional context**
Helps with #829

### DIFF
--- a/src/WorkflowCore/Interface/IWorkflowController.cs
+++ b/src/WorkflowCore/Interface/IWorkflowController.cs
@@ -34,6 +34,6 @@ namespace WorkflowCore.Interface
         /// <param name="workflowId"></param>
         /// <returns></returns>
         Task<bool> TerminateWorkflow(string workflowId);
-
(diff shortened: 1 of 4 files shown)


DevsAnon:
Add the current state this will not work as the failed execution pointer is still persisted and needs to be removed from DB as well. 
Trying to Look into ways on how to perform that, wouldn't mind some help

DevsAnon:
This is one change that could help with finalising the above
https://github.com/danielgerlag/workflow-core/pull/831

danielgerlag:
Why can't we suspend workflows when they produce errors instead of terminating them?

PR #831, opened 2021-05-14
Persist execution pointers

DevsAnon:
**Describe the change**
Always persist execution pointer changes when persisting a workflow


danielgerlag:
There are so many formatting changes here, it's impossible to find the actual change. Also please provide a more detailed description about what this is.

PR #834, opened 2021-05-17
support yaml definition of Outputs with Deep expression property

dwy12321:
**Describe the change**
Add Outputs assign with deep expression property.
As yaml definition workflow , i can not assign data to  custom data's sub sub property relevant to  workflow instance.
As following example,  when i assgin 0 to WorkflowCore.Sample03.MyDataClass sub sub property Value6, i get an exception.

DataType: WorkflowCore.Sample03.MyDataClass, WorkflowCore.Sample03
Steps:
-- Id: HelloWorldId
  Name: HelloWorld
  StepType: WorkflowCore.Sample03.Steps.HelloWorld, WorkflowCore.Sample03
  Outputs: 
    **myData.Value6: 0**

**Tests**

**Breaking change**
 compatibility with previous versions?


danielgerlag (commented):
Could we add some tests to demonstrate the new functionality working and also make sure it's not breaking indexed properties?  Could we also clean up the formatting?

dwy12321:
Hi  daniel,
I add WorkflowCore.Tests.YmalDefinition to test this change, In this test case, the data transmission is correct. Could you please help review this pr.

test/WorkflowCore.Tests.YmalDefinition/YmalDefinitionTest.cs
```diff
@@ -0,0 +1,46 @@
...
+using System.Linq;
+using System.Linq.Expressions;
+using System.Threading;
+using WorkflowCore.Interface;
+using WorkflowCore.Testing;
+using Xunit;
+
+namespace WorkflowCore.Tests.YmalDefinition
+{
+    public class YmalDefinitionTest:YamlWorkflowTest
```
danielgerlag, line 14: The changes you made are not YAML specific, they apply to the definition loader, regardless of serialization format.  I think we should add unit tests [https://github.com/danielgerlag/workflow-core/blob/master/test/WorkflowCore.UnitTests/Services/DefinitionStorage/DefinitionLoaderTests.cs](here). This particular test is more of an integration test, which we could add to the integration tests project.  I think the unit tests are more important in this case.

dwy12321: Thanks,good suggesttion, I have  added to  unit tests.

src/WorkflowCore.DSL/Services/DefinitionLoader.cs
```diff
@@ -217,17 +217,76 @@ private void AttachOutputs(StepSourceV1 source, Type dataType, Type stepType, Wo
...
                 // Check if our datatype has a matching property
-                var propertyInfo = dataType.GetProperty(output.Key);
-                if (propertyInfo != null)
+
+                PropertyInfo propertyInfo = null;
+                String[] paths = output.Key.Split('.');
+
+                bool hasAddOutput = false;
+
+                foreach (String propertyName in paths)
```
danielgerlag, line 229: This method is getting quite big and doing lots of things... could we split the nested property flattening into it's own method?

test/WorkflowCore.UnitTests/Services/DefinitionStorage/DefinitionLoaderTests.cs
```diff
@@ -43,6 +43,16 @@ public void ParseDefinition()
             A.CallTo(() => _registry.RegisterWorkflow(A<WorkflowDefinition>.That.Matches(MatchTestDefinition, ""))).MustHaveHappened();
         }
 
+        [Fact(DisplayName = "Should parse definition")]
+        public void ParseDefinitionPropertyDynamic()
```
danielgerlag, line 47: Should we also test that the nested output properties are getting mapped according to your change?

dwy12321: Ok , I add nested output properties in  test

src/WorkflowCore.DSL/Services/DefinitionLoader.cs
dwy12321: ok, split the nested property and direct property using AttachNestedOutput AttachDirectlyOutput

dwy12321:
Hi daniel,
Would you please help review this pull request?

PR #841, opened 2021-06-06
Add a new lock provider using MySql GET_LOCK and RELEASE_LOCK

paulose:
A new provider to allow distributed locking using MySql as the provider. It uses the GET_LOCK / RELEASE_LOCK functionality provided by MySql to acquire / release a lock. The code structure follows a similar structure like the lock provider for Sql Server

### DIFF
--- a/src/providers/WorkflowCore.LockProviders.MySQL/MySqlLockProvider.cs
+++ b/src/providers/WorkflowCore.LockProviders.MySQL/MySqlLockProvider.cs
@@ -1,0 +1,138 @@
+﻿using System;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+using WorkflowCore.Interface;
+using System.Data;
+using System.Collections.Generic;
+using System.Threading;
+using MySql.Data.MySqlClient;
+
+namespace WorkflowCore.LockProviders.MySQL
+{
+    public class MySqlLockProvider : IDistributedLockProvider
+    {
+        private const string Prefix = "wfc";
+
+        private readonly string _connectionString;
+        private readonly ILogger _logger;
+        private readonly Dictionary<string, MySqlConnection> _locks = new Dictionary<string, MySqlConnection>();
+        private readonly AutoResetEvent _mutex = new AutoResetEvent(true);
+
+        public MySqlLockProvider(string connectionString, ILoggerFactory logFactory)
+        {
+            _logger = logFactory.CreateLogger<MySqlLockProvider>();
+            var csb = new MySqlConnectionStringBuilder(connectionString);
+            csb.Pooling = true;
+            _connectionString = csb.ToString();
+        }
+
+
+        public async Task<bool> AcquireLock(string Id, CancellationToken cancellationToken)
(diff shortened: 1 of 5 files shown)


paulose:
Hi Daniel,

Please do have a look at this when you get the time. The PR is for a MySql based lock provider.

src/providers/WorkflowCore.LockProviders.MySQL/MySqlLockProvider.cs
```diff
@@ -0,0 +1,138 @@
...
+        }
+
+
+        public async Task<bool> AcquireLock(string Id, CancellationToken cancellationToken)
+        {
+            if (_mutex.WaitOne())
+            {
+                try
+                {
+                    var connection = new MySqlConnection(_connectionString);
```
danielgerlag, line 36: Does this come from a connection pool or is it establishing a fresh connection everytime?

src/providers/WorkflowCore.LockProviders.MySQL/MySqlLockProvider.cs
```diff
@@ -0,0 +1,138 @@
...
+                    await connection.OpenAsync(cancellationToken);
+                    try
+                    {
+                        var cmd = connection.CreateCommand();
+                        cmd.CommandText = $"SELECT GET_LOCK('{Prefix}:{Id}', 10)";
```
danielgerlag, line 41: Is that a lease time (10)?

src/providers/WorkflowCore.LockProviders.MySQL/MySqlLockProvider.cs
```diff
@@ -0,0 +1,138 @@
...
+using Microsoft.Extensions.Logging;
+using WorkflowCore.Interface;
+using System.Data;
+using System.Collections.Generic;
+using System.Threading;
+using MySql.Data.MySqlClient;
+
+namespace WorkflowCore.LockProviders.MySQL
+{
+    public class MySqlLockProvider : IDistributedLockProvider
```
danielgerlag, line 12: Could we add some tests against a real MySQL instance in the `WorkflowCore.Tests.MySQL` test assembly?
Let's also think about how lease time should be tested.

danielgerlag (changes requested):

src/providers/WorkflowCore.LockProviders.MySQL/MySqlLockProvider.cs
paulose: 10 is the timeout in seconds. It will wait for a maximum of 10 seconds for the query to give a result. If the timeout has exceeded, the query will return -1, which is handled as an exception below.

src/providers/WorkflowCore.LockProviders.MySQL/MySqlLockProvider.cs
paulose: It comes from a pool. You can see the code to set Pooling = true for connection string in the constructor.

Issue #844, opened 2021-06-09
Supply WorkflowOptions in WorkflowCore.Testing


Granjow:
**Is your feature request related to a problem? Please describe.**
When running unit tests with WorkflowCore.Testing, it is not possible to provide custom options to e.g. reduce the polling interval. The default interval of 10 s may cause tests to take a long time to complete.

**Describe the solution you'd like**
In WorkflowCore.Testing:

```csharp
    protected virtual void Setup()
    {
      IServiceCollection services = (IServiceCollection) new ServiceCollection();
      services.AddLogging();
      this.ConfigureServices(services);
      ServiceProvider serviceProvider = services.BuildServiceProvider();
      ServiceProviderServiceExtensions.GetService<ILoggerFactory>((IServiceProvider) serviceProvider);
      this.PersistenceProvider = ServiceProviderServiceExtensions.GetService<IPersistenceProvider>((IServiceProvider) serviceProvider);
      this.Host = ServiceProviderServiceExtensions.GetService<IWorkflowHost>((IServiceProvider) serviceProvider);
      this.Host.RegisterWorkflow<TWorkflow, TData>();
      this.Host.OnStepError += new StepErrorEventHandler(this.Host_OnStepError);
      this.ConfigureHost(); // <- add this
      this.Host.Start();
    }

    protected virtual void ConfigureHost()
    {
        // Could e.g. do that in the overriding method:
        // Host.Options.UsePollInterval(TimeSpan.FromMilliseconds(100));
    }
```

**Describe alternatives you've considered**

```csharp
Host.Stop();
Host.Options.UsePollInterval(TimeSpan.FromMilliseconds(100));
Host.Start();
```

This does not work because no events will be emitted anymore after stopping and starting the workflow host. This is due to bugs in Workflow Core. I will submit a separate PR to fix some of them.

wfcore-manager:
Closing as not planned — this has been open for several years without recent progress. Please open a new issue if this is still relevant on a current Workflow Core release.

PR #845, opened 2021-06-09
Fix lifecycle events when restarting Workflow Core

Granjow:
**Describe the change**

LifeCycleEventPublisher does not add any events anymore when it is stopped and started again. This PR fixes this behaviour and also attaches lifecycle events again which are removed when stopping the host.

See code

**Tests**
Publisher is covered, Host is not covered.

**Breaking change**
None.

**Additional context**
Related to #844, but does not completely fix the issues described there.


danielgerlag (approved):

Granjow:
Thanks for merging!

PR #851, opened 2021-06-23
Use lock on event subscription

VKAlwaysWin:
Maybe related to https://github.com/danielgerlag/workflow-core/pull/782

Hi @danielgerlag ,

Let me share some scenario:
Input: Workflow with multiple Parallel steps and in each of it WaitFor tasks.
At some point we are facing next situation:

1. In custom step we are sending some work to process (in our case it is a queue).
2. After that we are adding WaitFor to subscribe on event (and it is working well). Job is returned back even if we didn't subscribe yet for it with WaitFor.
This case if handled in:
But Event is not locked here : https://github.com/danielgerlag/workflow-core/blob/master/src/WorkflowCore/Services/BackgroundTasks/WorkflowConsumer.cs#L94

3. After subscription we are reseting event back to unprocessed state. Which is ok.
4. Poller on line: https://github.com/danielgerlag/workflow-core/blob/master/src/WorkflowCore/Services/BackgroundTasks/RunnablePoller.cs#L90 . Will pickup this event because it is unprocessed.
5. But workflow can still be locked, because one of the parallel steps are doing their job. Event Consumer again won't process event and Poller will put Event to greylisted list till forever for one instance.
https://github.com/danielgerlag/workflow-core/blob/master/src/WorkflowCore/Services/BackgroundTasks/RunnablePoller.cs#L96

(will be in greylist forever)

### DIFF
--- a/src/WorkflowCore/Services/BackgroundTasks/EventConsumer.cs
+++ b/src/WorkflowCore/Services/BackgroundTasks/EventConsumer.cs
@@ -74,7 +74,13 @@ namespace WorkflowCore.Services.BackgroundTasks
                         complete = complete && await SeedSubscription(evt, sub, toQueue, cancellationToken);
 
                     if (complete)
+                    {
                         await _eventRepository.MarkEventProcessed(itemId, cancellationToken);
+                    }
+                    else
+                    {
+                        _greylist.Remove($"evt:{evt.Id}");
+                    }
 
                     foreach (var eventId in toQueue)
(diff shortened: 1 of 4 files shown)


VKAlwaysWin:
@danielgerlag please let me know if something doesn't make sense for you.

danielgerlag:
@VKAlwaysWin I merged this, but the test you added is failing.

VKAlwaysWin:
@danielgerlag looking on this. Will post a new PR with fix

Issue #542, opened 2020-03-31
Activities on multiple workflow instances


pandene:
When you have multiple instances of the same workflow containing Activities the GetPendingActivity method has no way to assosiate this with a specific workflow instance but instead find the first open subscription( created by all the Activity instances) in the repo
Maybe I do not understand this correctly but an activity cannot happen if it is not associated with a workflow instance otherwise this just becomes a random activity.

To illustrate my point:
If you take sample18(approval of spending). If you create two instances of the workflow there is no way to say which one you approve unless you can specify a workflow.

Things will also slow down as the amount of events grow in the repo:
            foreach (var eventId in await _eventRepository.GetEvents(sub.EventName, sub.EventKey, sub.SubscribeAsOf))
This line will go through all events in the repo(can become millions) with the same name to see if it must be executed

danielgerlag:
You can make a compound key with some unique ID from the workflow instance as the activity name.

ripvannwinkler:
> You can make a compound key with some unique ID from the workflow instance as the activity name.

How, exactly? The example code doesn't expose any way to get the workflow instance id.

	class ActivityWorkflow : IWorkflow<MyData>
	{
		public string Id => "activity-sample";
		public int Version => 1;

		public void Build(IWorkflowBuilder<MyData> builder)
		{
			builder
				.Activity("get-approval", (data) => data.Request)
					.Output(data => data.ApprovedBy, step => step.Result)
				.Then<CustomMessage>()
					.Input(step => step.Message, data => "Approved by " + data.ApprovedBy)
				.Then<GoodbyeWorld>();
		}
	}

How can I make "get-approval" reference the specific workflow instance here? What am I missing?

PR #864, opened 2021-07-07
Overload GetPendingActivity with workflowId parameter. Solves #542 issue.

MathSee:
**Describe the change**
An overload on the `GetPendingActivity `on `ActivityController` to include `workflowId ` parameter, this overload allows the host to get only the pending activities from a specific workflow. Helping to solve the issue #542  

It's only a overload, follows the same implementation that the master branch, but now help the host to get pending activities with a new filter, the workflowId.

**Tests**
Yes

**Breaking change**
No

**Additional context**
This change is not supported only on Redis provider. I couldn't implements this overload for Redis.


wfcore-manager:
Closing as stale (no updates since 2021). Please reopen with a refreshed branch if this is still needed.

PR #870, opened 2021-07-15
Do not do polling If Events or Workflows disabled

VKAlwaysWin:
Added:
if (_options.EnableWorkflows)
if (_options.EnableEvents)


danielgerlag (approved):

danielgerlag:
What if someone wants a configuration where one node does the polling and another does the processing?

VKAlwaysWin:
@danielgerlag that sounds like edge case for me, much more possible that someone is disabling Events but Polling service will conitnue do locking. 
If you want I can prepare another PR for this edge case scenario, but I don't think that it will affect current fix.

danielgerlag:
What makes it an edge case? Seems like a legitimate way to configure things, I am pretty sure people are already doing this today.

PR #884, opened 2021-08-03
Allow to disable events publisher

VKAlwaysWin:
**Describe the change**
Allow to disable LifeCycleEventsPublisher

**Breaking change**
No breaking changes.

### DIFF
--- a/src/WorkflowCore/Services/ErrorHandlers/SuspendHandler.cs
+++ b/src/WorkflowCore/Services/ErrorHandlers/SuspendHandler.cs
@@ -29,6 +29,8 @@ namespace WorkflowCore.Services.ErrorHandlers
                 WorkflowDefinitionId = workflow.WorkflowDefinitionId,
                 Version = workflow.Version
             });
+
+            step.PrimeForRetry(pointer);
         }
     }
 }
--- a/src/WorkflowCore/Services/ErrorHandlers/SuspendHandler.cs
+++ b/src/WorkflowCore/Services/ErrorHandlers/SuspendHandler.cs
@@ -21,7 +21,7 @@ namespace WorkflowCore.Services.ErrorHandlers
         public void Handle(WorkflowInstance workflow, WorkflowDefinition def, ExecutionPointer pointer, WorkflowStep step, Exception exception, Queue<ExecutionPointer> bubbleUpQueue)
         {
             workflow.Status = WorkflowStatus.Suspended;
-            _eventPublisher?.PublishNotification(new WorkflowSuspended
+            _eventPublisher.PublishNotification(new WorkflowSuspended
             {
                 EventTimeUtc = _datetimeProvider.UtcNow,
                 Reference = workflow.Reference,
--- a/src/WorkflowCore/Services/ErrorHandlers/SuspendHandler.cs
+++ b/src/WorkflowCore/Services/ErrorHandlers/SuspendHandler.cs
@@ -21,7 +21,7 @@ namespace WorkflowCore.Services.ErrorHandlers
         public void Handle(WorkflowInstance workflow, WorkflowDefinition def, ExecutionPointer pointer, WorkflowStep step, Exception exception, Queue<ExecutionPointer> bubbleUpQueue)
         {
             workflow.Status = WorkflowStatus.Suspended;
-            _eventPublisher.PublishNotification(new WorkflowSuspended
+            _eventPublisher?.PublishNotification(new WorkflowSuspended
             {
                 EventTimeUtc = _datetimeProvider.UtcNow,
                 Reference = workflow.Reference,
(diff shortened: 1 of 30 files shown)


danielgerlag (commented):
What problem is this trying to solve?

VKAlwaysWin:
@danielgerlag Performance mainly. Those events used only in case if you have OnLifeCycleEvent and in all other cases it is not used. Please correct me if I wrong here.

In case of really big load 10+ workflows per second with a lot of steps GetConsumingEnumerable for BlockingCollection is taking couple of percents (~ 3.5% in my case) of execution time. 

And this PR allowing to save couple of ms per workflow if you don't need to track LifeCycle events. Plus it shouldn't affect current default settings and implementation.

Does that make sense ?

src/WorkflowCore/Services/ErrorHandlers/SuspendHandler.cs
```diff
@@ -21,7 +21,7 @@ public SuspendHandler(ILifeCycleEventPublisher eventPublisher, IDateTimeProvider
         public void Handle(WorkflowInstance workflow, WorkflowDefinition def, ExecutionPointer pointer, WorkflowStep step, Exception exception, Queue<ExecutionPointer> bubbleUpQueue)
         {
             workflow.Status = WorkflowStatus.Suspended;
-            _eventPublisher.PublishNotification(new WorkflowSuspended
+            _eventPublisher?.PublishNotification(new WorkflowSuspended
```
danielgerlag, line 24: The fact that the publisher can be disabled is leaking into where it is consumed, this creates inappropriate coupling.

src/WorkflowCore/Services/ErrorHandlers/TerminateHandler.cs
```diff
@@ -23,7 +23,7 @@ public void Handle(WorkflowInstance workflow, WorkflowDefinition def, ExecutionP
             workflow.Status = WorkflowStatus.Terminated;
             workflow.CompleteTime = _dateTimeProvider.UtcNow;
 
-            _eventPublisher.PublishNotification(new WorkflowTerminated
+            _eventPublisher?.PublishNotification(new WorkflowTerminated
```
danielgerlag, line 26: same here

src/WorkflowCore/Services/ExecutionResultProcessor.cs
```diff
@@ -77,7 +77,7 @@ public void ProcessExecutionResult(WorkflowInstance workflow, WorkflowDefinition
                     subsequent.Active = true;
                 }
 
-                _eventPublisher.PublishNotification(new StepCompleted
+                _eventPublisher?.PublishNotification(new StepCompleted
```
danielgerlag, line 80: same here

src/WorkflowCore/Services/ExecutionResultProcessor.cs
```diff
@@ -102,7 +102,7 @@ public void ProcessExecutionResult(WorkflowInstance workflow, WorkflowDefinition
 
         public void HandleStepException(WorkflowInstance workflow, WorkflowDefinition def, ExecutionPointer pointer, WorkflowStep step, Exception exception)
         {
-            _eventPublisher.PublishNotification(new WorkflowError
+            _eventPublisher?.PublishNotification(new WorkflowError
```
danielgerlag, line 105: same here

src/WorkflowCore/Services/WorkflowExecutor.cs
```diff
@@ -123,7 +123,7 @@ private bool InitializeStep(WorkflowInstance workflow, WorkflowStep step, Workfl
             if (pointer.Status != PointerStatus.Running)
             {
                 pointer.Status = PointerStatus.Running;
-                _publisher.PublishNotification(new StepStarted
+                _publisher?.PublishNotification(new StepStarted
```
danielgerlag, line 126: same here

src/WorkflowCore/Services/WorkflowExecutor.cs
```diff
@@ -264,7 +264,7 @@ private async Task DetermineNextExecutionTime(WorkflowInstance workflow, Workflo
                 await middlewareRunner.RunPostMiddleware(workflow, def);
             }
 
-            _publisher.PublishNotification(new WorkflowCompleted
+            _publisher?.PublishNotification(new WorkflowCompleted
```
danielgerlag, line 267: same here

src/WorkflowCore/Services/ErrorHandlers/SuspendHandler.cs
danielgerlag: Why not just put the check in the publisher?

VKAlwaysWin:
@danielgerlag added check before adding into outbox. Will that work for you ?
Also added unit test to cover this use case and improved existing ones.

VKAlwaysWin:
@danielgerlag any chance to get this merged ?

PR #908, opened 2021-09-04
Align MaxLength of EventName and EventKey among entities.

### LINKED ISSUE #898: PersistedExecutionPointer's EventName and EventKey are shorter than PersistedEvent's ones
**Describe the bug**
EventName and EventKey properties of PersistedExecutionPointer are attributed as MaxLength(100), same properties of PersistedEvent are attributed as MaxLength(200).
When WFC attempts to save event name and event key to ExecutionPointer in WaitFor() operation that are the same as in PublishEvent() and exceed 100 symbols execution fails with error 'SQL Server Error : String or binary data would be truncated'.
At that PublishEvent is successful.

**Expected behavior**
PublishEvent() and WaitFor() should behave consistently so MaxLength on both PersistedExecutionPointer and PersistedEvent should be aligned and be at least 200 (taking into account production data).

vladimir-kovalyuk:
Made MaxLength of EventName and EventKey properties the same as in PersistedEvent.
Added EF Migration.

fixes #898

### DIFF
--- a/src/providers/WorkflowCore.Persistence.EntityFramework/Models/PersistedExecutionPointer.cs
+++ b/src/providers/WorkflowCore.Persistence.EntityFramework/Models/PersistedExecutionPointer.cs
@@ -32,10 +32,10 @@ namespace WorkflowCore.Persistence.EntityFramework.Models
 
         public DateTime? EndTime { get; set; }
 
-        [MaxLength(100)]
+        [MaxLength(200)]
         public string EventName { get; set; }
 
-        [MaxLength(100)]
+        [MaxLength(200)]
         public string EventKey { get; set; }
 
         public bool EventPublished { get; set; }
--- a/src/providers/WorkflowCore.Persistence.PostgreSQL/Migrations/20210907073438_Align_EventKey_Length.Designer.cs
+++ b/src/providers/WorkflowCore.Persistence.PostgreSQL/Migrations/20210907073438_Align_EventKey_Length.Designer.cs
@@ -2,6 +2,7 @@
 using System;
 using Microsoft.EntityFrameworkCore;
 using Microsoft.EntityFrameworkCore.Infrastructure;
+using Microsoft.EntityFrameworkCore.Migrations;
 using Microsoft.EntityFrameworkCore.Storage.ValueConversion;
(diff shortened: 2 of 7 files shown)


vladimir-kovalyuk:
https://github.com/danielgerlag/workflow-core/issues/898

src/providers/WorkflowCore.Persistence.EntityFramework/Models/PersistedExecutionPointer.cs
```diff
@@ -32,10 +32,10 @@ public class PersistedExecutionPointer
 
         public DateTime? EndTime { get; set; }
 
-        [MaxLength(100)]
+        [MaxLength(200)]
         public string EventName { get; set; }
 
-        [MaxLength(100)]
+        [MaxLength(200)]
```
danielgerlag, line 38: We probably need to create a migration for Postgres too?

danielgerlag (commented):
You mentioned in your comment that you had to update to .net 5? 
Which project was that?

vladimir-kovalyuk:
> You mentioned in your comment that you had to update to .net 5?
> Which project was that?
I was unable to run dotnet ef commands.
It seems that this line
`
    <DotNetCliToolReference Include="Microsoft.EntityFrameworkCore.Tools.DotNet" Version="1.0.0-msbuild3-final" />
`
is no longer working, at least in my environment.
So I had to target net5.0 in WorkflowCore.Persistence.SqlServer project temporarily to enable dotnet ef commands and rolled project changes back then.
I tried to install/update Microsoft.EntityFrameworkCore.Tools.DotNet library to 2.0.3 version and obtained another portion of issues, some of them were about test WorkflowCore.Tests.SqlServer project.

src/providers/WorkflowCore.Persistence.EntityFramework/Models/PersistedExecutionPointer.cs
vladimir-kovalyuk: Added migration for PostgreSql

vladimir-kovalyuk: Not sure about MySql if it requires local MySql server connection in order to create migration.

PR #913, opened 2021-09-14
Fix SQL Server tests

SergiiKram:
The repository` microsoft/mssql-server-linux` is no longer maintained and the container image cannot be pulled.
The `mcr.microsoft.com/mssql/server` should be used instead.

See https://cloudblogs.microsoft.com/sqlserver/2021/05/20/microsoft-sql-server-linux-based-container-images-to-be-available-only-from-the-microsoft-container-registry/ for details.

Without the fix test fail with the following error:
"`pull access denied for microsoft/mssql-server-linux, repository does not exist or may require 'docker login': denied: requested access to the resource is denied`".


SergiiKram:
Hi @danielgerlag, I want to bring your attention to this PR, it fixes completely broken tests for MS SQL Server.

PR #914, opened 2021-09-14
Fix DynamoDB tests

SergiiKram:
The recent `amazon/dynamodb-local` images introduce breaking change and require dummy credentials, see https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/DynamoDBLocal.DownloadingAndRunning.html.

Without the fix, tests fail with the timeout error in `DockerSetup.StartContainer` method but the real error is "`Unable to get IAM security credentials from EC2 Instance Metadata Service.`" in the `DynamoDbDockerSetup.TestReady` method.


SergiiKram:
Hi @danielgerlag, yet another PR to fix completely broken tests, this time for DynamoDB.

Issue #828, opened 2021-05-06
Idempotency when creating workflows


creyke:
**Is your feature request related to a problem? Please describe.**
As a client of an `IWorkflowHost` host I can start a workflow instance with a reference, which could be my idempotency key, but I cannot check on the status of this workflow instance unless I successfully receive the workflow id back. I am also able to start more than one workflow instance with the same reference. This means that in some cases I am unable to validate that my initial request was unsuccessful and I risk starting multiple workflows where I only intended to start one.

**Describe the solution you'd like**
The ability to define the workflow id as a client to an idempotent implementation of `IWorkflowHost.StartWorkflow()`. In the event this workflow already exists, this call would not kick off another workflow.

**Describe alternatives you've considered**
Returning the workflow id, although this doesn't solve the idempotency issues.

**Additional context**
N/A

danielgerlag:
Not sure this is something the workflow engine should take responsibility for, have you considered implementing idempotency checks in the calling process?

creyke:
Hi Daniel thank you for the quick response. The Web API in front of the ‘IWorkflowHost’ is already idempotent but this will not guarantee at most once instancing of a workflow, as if the process is terminated before the response is received by the client of the API, the client must assume the request was not received and duplicate the request.

vladimir-kovalyuk:
I need to prevent multiple active instances of some workflows. To achieve this I didn't find anything better than the following:
        public bool IsWorkflowActivated(string workflowId)
        {
            var runnableInstances = _workflowRepository.GetRunnableInstances(DateTime.Now);
            var workflowInstances = _workflowRepository.GetWorkflowInstances(runnableInstances);
            return workflowInstances.Any(wi => wi.WorkflowDefinitionId == workflowId);
        }
My understanding is that this code makes huge load on the persistence tier and CPU as well (as long as it jsondeserializes all the instances).
Is there any better approach? Or in case of 'no' maybe it's worth considering adding such method to API?

danielgerlag:
You could store the workflow instance ID and your reference in your own storage and check against that?

creyke:
Would that not be a two phase commit? Paradoxically that would arguably
require a workflow to orchestrate. Would it be possible to supply the
workflow id when creating it, and then it cannot be created twice as
storage will be the single source of truth?

vladimir-kovalyuk:
> You could store the workflow instance ID and your reference in your own storage and check against that?

[...]

PR #921, opened 2021-09-22
Make StartWorkflow idempotent

SergiiKram:
**The change**
The optional correlation id parameter is added with a unique constraint in the database.
So instead of creating a duplicate workflow instance, we can get the id of the previously created one.

Correlation id can also be used to find the workflow.

So now we can safely retry workflow creation and have an exactly-once guarantee.
Also, in scenarios where we create a workflow to process document, user, external event, etc., we can derive correlation id from these entities and later find the matching workflow if we failed to persist the returned workflow id.

Example:
```
// correlationId is provided by the calling client
var id = await _workflowController.StartWorkflow(workflowId, correlationId: correlationId);

// Subsequent call does nothing and returns same id.
var id2 = await _workflowController.StartWorkflow(workflowId, correlationId: correlationId);
```

**The design**
The implementation is quite trivial and standard for such cases. A unique constraint is created for the CorrelationId field in the database. So the insert would fail in case of a retry with the same correlation id. The relevant exception is caught in the `WorkflowController` and the id of the already created workflow is returned.

**Tests**
The change is well covered with tests to ensure both backward compatibility and implementation correctness.

**Breaking changes**
The WorkflowCore.Persistence.MySQL project was upgraded to netstandard2.1. However, this can be discussed and reverted if breaking change is not welcomed.

**Additional considerations**
I didn't implement the support for a few persistence providers due to the absence of tests and to get feedback sooner: RavenDB, Dynamo, Cosmos DB, and Redis.
The NotImplementedException will be thrown when someone will try to create a workflow and the correlation id will be passed.

**Next steps**
- [ ] Update documentation.
- [ ] Update WorkflowCore.WebAPI.

**Relevant issues**
The feature was previously requested in the community, e.g. https://github.com/danielgerlag/workflow-core/issues/828

### DIFF
--- a/src/providers/WorkflowCore.Persistence.RavenDB/Services/RavendbPersistenceProvider.cs
+++ b/src/providers/WorkflowCore.Persistence.RavenDB/Services/RavendbPersistenceProvider.cs
@@ -41,6 +41,11 @@ namespace WorkflowCore.Persistence.RavenDB.Services
 
 		public async Task<string> CreateNewWorkflow(WorkflowInstance workflow, CancellationToken cancellationToken = default)
 		{
+			if (workflow.CorrelationId != null)
+			{
+				throw new NotImplementedException();
+			}
+
 			using (var session = _database.OpenAsyncSession())
 			{
 				await session.StoreAsync(workflow, cancellationToken);
@@ -91,6 +96,11 @@ namespace WorkflowCore.Persistence.RavenDB.Services
 				return result;
 			}
 		}
+
+		public Task<WorkflowInstance> GetWorkflowInstanceByCorrelationId(string correlationId, CancellationToken cancellationToken = default)
+        {
+			throw new NotImplementedException();
+        }
 
 		public async Task<IEnumerable<WorkflowInstance>> GetWorkflowInstances(IEnumerable<string> ids, CancellationToken cancellationToken = default)
 		{
--- a/src/WorkflowCore/Exceptions/WorkflowExistsException.cs
+++ b/src/WorkflowCore/Exceptions/WorkflowExistsException.cs
@@ -1,0 +1,23 @@
+﻿using System;
+using System.Runtime.Serialization;
+
+namespace WorkflowCore.Exceptions
+{
+    public class WorkflowExistsException : Exception
+    {
+        public WorkflowExistsException()
+            : base("Workflow with provided correlation id already exists.")
+        {
+        }
+
+        protected WorkflowExistsException(SerializationInfo info, StreamingContext context)
+            : base(info, context)
+        {
+        }
+
+        public WorkflowExistsException(Exception innerException)
+            : base("Workflow with provided correlation id already exists.", innerException)
+        {
+        }
+    }
+}
--- a/src/WorkflowCore/Interface/IWorkflowController.cs
+++ b/src/WorkflowCore/Interface/IWorkflowController.cs
@@ -5,10 +5,10 @@ namespace WorkflowCore.Interface
 {
     public interface IWorkflowController
     {
-        Task<string> StartWorkflow(string workflowId, object data = null, string reference=null);
-        Task<string> StartWorkflow(string workflowId, int? version, object data = null, string reference=null);
-        Task<string> StartWorkflow<TData>(string workflowId, TData data = null, string reference=null) where TData : class, new();
-        Task<string> StartWorkflow<TData>(string workflowId, int? version, TData data = null, string reference=null) where TData : class, new();
+        Task<string> StartWorkflow(string workflowId, object data = null, string reference=null, string correlationId = null);
+        Task<string> StartWorkflow(string workflowId, int? version, object data = null, string reference=null, string correlationId = null);
+        Task<string> StartWorkflow<TData>(string workflowId, TData data = null, string reference=null, string correlationId = null) where TData : class, new();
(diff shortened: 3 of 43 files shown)


SergiiKram:
Hi @danielgerlag,

I would appreciate it if you consider accepting this pull request as a feature proposal to add idempotency support for workflow creation.
It's a show stopper for my project and I guess it would be a valuable feature for many users who are concerned about the resiliency of their solutions.

src/providers/WorkflowCore.Persistence.RavenDB/Services/RavendbPersistenceProvider.cs
```diff
@@ -41,6 +41,11 @@ static void CreateIndexes(RavendbPersistenceProvider instance)
 
 		public async Task<string> CreateNewWorkflow(WorkflowInstance workflow, CancellationToken cancellationToken = default)
 		{
+			if (workflow.CorrelationId != null)
```
vladimir-kovalyuk, line 44: Wouldn't it break existing solutions based on RavenDB?

vladimir-kovalyuk:
Have you reviewed your PR on your own?
A lot of changes in migration code, .csproj files,  NotImplementedException-s

src/providers/WorkflowCore.Persistence.RavenDB/Services/RavendbPersistenceProvider.cs
SergiiKram: No, since any existing solution does not pass the CorrelationId parameter (i.e. CorrelationId would be null), so nothing would change and behavior would be the same as before.

If someone would try to get benefits of the CorrelationId with RavenDB persistence only then they would see an error encouraging to contribute with tests first and then similar change - add unique constraint in the database.

SergiiKram:
Thanks for the review, Vladimir @vladimir-kovalyuk.

> A lot of changes in migration 

The change is trivial - add a new (optional) column/field to the database with the unique constraint.
And that obviously means we need to introduce migration to the database schema.
I agree that the migration code is quite verbose and duplicated for each persistence provider, but it was automatically generated by Entity Framework, so it's quite safe and straightforward.

> .csproj files

There is only one file that changed: WorkflowCore.Persistence.MySQL.csproj. I dropped netstandard2.0 support since it creates some compatibility issues going forward and other persistence providers already support netstandard2.1 only (SQL Server, PostgreSQL, SQLite).

> NotImplementedException-s

As I mentioned, I didn't implement changes for few providers mostly because there are no tests for them at all. The support can be added later time in the same fashion if there is a need for anybody.
Again, nothing would break for existing users of RavenDB e.g. if they would not use the new feature.

SergiiKram:
Some good comments here:
https://github.com/danielgerlag/workflow-core/issues/828#issuecomment-926464528

danielgerlag:
Could you not achieve this with the library as is by using a [TransactionScope](https://docs.microsoft.com/en-us/dotnet/api/system.transactions.transactionscope?view=net-5.0)?

SergiiKram:
> Could you not achieve this with the library as is by using a [TransactionScope](https://docs.microsoft.com/en-us/dotnet/api/system.transactions.transactionscope?view=net-5.0)?


Not sure I get your suggestion right, Daniel. So thinking out loud.
I presume the idea is to use MS SQL Server Distributed Transaction Manager to have a distributed transaction across two databases (or two individual transactions in the same DB?).

I see a few issues here:

- The solution is not portable and we do not use MS SQL Server.
- It looks like an attempt to use 2 phase commit in a distributed system and I would prefer to use the eventual consistency approach here.
- Additional complexity and points of failure for the trivial use case.

From my perspective idempotency is an important quality for the framework to build resilient and high-load distributed systems and I'd be happy to contribute back to the OSS community.

danielgerlag:
No, it is not about using a distributed transaction across 2 databases or even using MsSQL. You can use the transaction scope to create a single transaction in the same DB on any DB that supports transactions. You could add your own table to the workflow DB and write to it as the first step in the transaction, then start the workflow, if the workflow fails to start, the entry will be rolled back... no 2 phase commit is needed.

danielgerlag:
Another approach would be to leverage the Reference field that is already on the startWorkflow overloads, you just need a unique constraint on that field in your particular DB.

SergiiKram:
Ah, I see, thanks.

The first thing I did I investigated the Reference field)) And indeed, adding a unique constraint to it would solve the issue.
Unfortunately, that would mean a breaking change for existing users.

We can maintain a custom fork with any implementation, obviously, but I would consider it as a backup option if the suggested approach is not aligned with your vision of the Workflow Core evolution.

danielgerlag:
You don't have to maintain a fork to use your own customized persistence provider.

SergiiKram:
Thanks for the suggestion, Daniel.
A custom persistence provider can be a compromise here.

Anyway, please close this PR if you are not comfortable merging it at this time.
Speaking of my roadmap - I was planning to port the fix to Cosmos DB (Microsoft has published the Linux container so we can add integration tests.)

SergiiKram:
I've create a nuget for PosgreSQL with `Reference` as an idempotency key.
See https://github.com/SergiiKram/workflow-core#idempotent-workflow-core

The diff https://github.com/SergiiKram/workflow-core/commit/b1b54c9c3055ea439f6fb67c14a2414cd0a7097d

I'm closing this PR, @danielgerlag feel free to ping me if you would be interested in porting idempotency fixes in the future.
Thanks guys!

PR #960, opened 2021-12-01
Add (auto) Id property to `WorkflowCore.Models.ExecutionError` so that it can easily be deserialized.

brettwinters:
**Describe the change**
Add (auto) Id property to `WorkflowCore.Models.ExecutionError` so that it can easily be deserialized using the MongoDb driver.

### DIFF
--- a/src/WorkflowCore/Models/ExecutionError.cs
+++ b/src/WorkflowCore/Models/ExecutionError.cs
@@ -4,7 +4,9 @@ namespace WorkflowCore.Models
 {
     public class ExecutionError
     {
-        public DateTime ErrorTime { get; set; }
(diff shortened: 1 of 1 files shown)


brettwinters:
I've just found out this is a breaking change since any previously inserted `ExecutionError` records have an _id field of type `ObjectId("xxxxxxxxxxxxx")` which cannot be deserialized into a string. 

I've solved this issue by registering a `IgnoreExtraElementsConvention` convention pack on `Startup` as follows

```c#
ConventionRegistry.Register(
    name: $"Ignore _id in {nameof(ExecutionError)}", 
    conventions: new ConventionPack() { new IgnoreExtraElementsConvention(true) }, 
    filter: t => t == typeof(ExecutionError)
);
```

PR #969, opened 2021-12-18
Resolve #957, E11000 duplicate key error

### LINKED ISSUE #957: Deferred execution duplicate index (MongoDB) error
**Describe the bug**
When making use of the deferred execution (as demo'd in https://github.com/danielgerlag/workflow-core/tree/master/src/samples/WorkflowCore.Sample05), 

e.g.
`return ExecutionResult.Sleep(pollingDelay, new object());`

I notice the following error (log):
A write operation resulted in an error. WriteError: { Category : "DuplicateKey", Code : 11000, Message : "E11000 duplicate key error collection: foobar.wfc.scheduled_commands index: idx_key dup key: { CommandName: "ProcessWorkflow", Data: "619cefda1e173f3b8cc532be" }" }.

**To Reproduce**
Run multiple hosts (6 at the moment) using MongoDB persistence; RedisLocking, RediseQueues and RedisEventHub.  Defer the execution for an activity (as instructed at https://github.com/danielgerlag/workflow-core/tree/master/src/samples/WorkflowCore.Sample05).

When the activity is deferred, an error is logged out every 10 seconds.

**Expected behavior**
Activity should be delayed/deferred, execute and no errors be highlighted in logs

**Additional context**
running WorkflowCore 3.6.1

potatopeelings:
**Describe the change**
Resolve #957. `InsertOneAsync` will throw a `MongoWriteException` and not a `MongoBulkWriteException`. Hence the original catch would have never triggered.

- Changed `MongoBulkWriteException` to `MongoWriteException` and changed the check on the `WriteErrors` property to a check on `WriteError`. 
- Also changed the `.` to `?.` (if the error is a WriteConcernError, the `WriteError` property won't be set)

**Tests**
I couldn't find any automated tests to modify in WorkflowCore.Tests.MongoDB, so I manually tested it (i.e. made sure the `MongoWriteException` is triggered, caught and ignored)

**Breaking change**
NOT a breaking change

**Additional context**
None

### DIFF
--- a/src/providers/WorkflowCore.Persistence.MongoDB/Services/MongoPersistenceProvider.cs
+++ b/src/providers/WorkflowCore.Persistence.MongoDB/Services/MongoPersistenceProvider.cs
@@ -321,6 +321,12 @@ namespace WorkflowCore.Persistence.MongoDB.Services
                      return;
                 throw;
             }
+            catch (MongoBulkWriteException ex)
+            {
+                if (ex.WriteErrors.All(x => x.Category == ServerErrorCategory.DuplicateKey))
+                    return;
+                throw;
+            }
         }
 
         public async Task ProcessCommands(DateTimeOffset asOf, Func<ScheduledCommand, Task> action, CancellationToken cancellationToken = default)
--- a/src/providers/WorkflowCore.Persistence.MongoDB/Services/MongoPersistenceProvider.cs
+++ b/src/providers/WorkflowCore.Persistence.MongoDB/Services/MongoPersistenceProvider.cs
@@ -315,9 +315,9 @@ namespace WorkflowCore.Persistence.MongoDB.Services
             {
                 await ScheduledCommands.InsertOneAsync(command);
             }
-            catch (MongoBulkWriteException ex)
+            catch (MongoWriteException ex)
             {
-                if (ex.WriteErrors.All(x => x.Category == ServerErrorCategory.DuplicateKey))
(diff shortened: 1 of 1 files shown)


src/providers/WorkflowCore.Persistence.MongoDB/Services/MongoPersistenceProvider.cs
```diff
@@ -315,9 +315,9 @@ public async Task ScheduleCommand(ScheduledCommand command)
             {
                 await ScheduledCommands.InsertOneAsync(command);
             }
-            catch (MongoBulkWriteException ex)
+            catch (MongoWriteException ex)
```
danielgerlag, line 318: Can we check for both types?

potatopeelings: Future proofing? Done.

The Mongo C# driver converts any `MongoBulkWriteException` thrown by `InsertOneAsync` into a `MongoWriteException` (https://github.com/mongodb/mongo-csharp-driver/blob/master/src/MongoDB.Driver/MongoCollectionBase.cs#L501). Exceptions thrown are not documented as part of the method API, so that can change.

danielgerlag: Actually, I think it did change at some point... when I was debugging with an older version I get MongoBulkWriteException.

danielgerlag (approved):

PR #980, opened 2022-01-08
Removed hard coded Max Concurrent Items in Event Consumer. 

dekrevedko:
That significantly impact performance in a lot of events scenario.


chrisbertrandtoca:
I agree this is a massive bottleneck when doing Event based flows. @danielgerlag is this something we can get in? If you'd prefer I can use the MaxConcurrentFlows variable that has been used for Workflows? Happy to PR that in, if we can get it in?

PR #1034, opened 2022-05-12
Cannot start WorkflowHost with MySql on NET6, DbContext injection missing

lfbertucci:
Hello Daniel,

After the changes on WorkflowCore to run with NET6, MySQL provider stopped working, throwing an error when trying to start de WorkflowCore Host.

The issue only occurs when using MySQL provider with NET6 projects.

After investigating, I found a condition in "MysqlContext" class that validate only NetStandard.

So I'm making a PR and adding the NET6 verification.

Thanks


lfbertucci:
I noticed that are some errors occurring at unit tests at master...
Do I need to do something else to resolve them?

PR #1052, opened 2022-06-22
Add ProceedOnCancel to WorkflowCore.DSL definition loader

lfbertucci:
Hi Daniel,

during my research to find a way to create a step with deadline, I found some post that suggest to use a .Parallel() with .WaitFor() on one branch and .Sleep() on another branch, and use .CancelCondition(<condition>, **true**) to the Parallel.

I could simulate my scenario and it works perfectly with FluentAPI, but I'm using the JSON Loader extension on WorkflowCore.DSL, and this loader not have how to pass the property "ProceedOnCancel".

So I added property "ProceedOnCancel" to StepSourceV1 and pass it value to the targetStep on ConvertStep at definitionLoader. And also change some tests to contemple this property.

Thank you for this amazing workflow engine.


lfbertucci:
Hello Daniel,

Any point about that PR?

Could you review it, please?

thanks you so much

Issue #1041, opened 2022-06-10
.net standard 2.0 support


uuggene:
Is it possible to support new table ScheduledCommand in netstandard2.0?

I found that WorkflowCore.Persistence.EntityFramework requires netstandard2.1 starting from version 3.6.0.

But I a lot of other libraries still support netstandard2.0.
This change is very critical, because we can't use this library in .net 4.6 projects anymore.

I tried to find why WorkflowCore.Persistence.EntityFramework requires netstandard2.1, and there are only 2 changes in one file to support netstandard2.0.
![D78AuQhoWc](https://user-images.githubusercontent.com/11631048/173063397-190d0e27-5372-420e-b49c-5f44192a4027.png)

danielgerlag:
EFCore dropped support for netstandard2.0 in v5

uuggene:
It's still possible to use EFCore 3.1.* as you do in project WorkflowCore.Persistence.EntityFramework:

    <PackageReference Include="Microsoft.EntityFrameworkCore" Version="3.1.2" />
    <PackageReference Include="Microsoft.EntityFrameworkCore.Relational" Version="3.1.2" />

PR #1057, opened 2022-06-23
Support netstandard2.0 in WorkflowCore.Persistence.EntityFramework #1041

uuggene:
There's no need to limit with netstandard2.1 as Project references EFCore 3.1.2.


genna:
Hi @danielgerlag. Is it possible that the current PR could be merged? It's blocking us from upgrading to the latest version of the workflow-core within our .NET Framework app

Issue #1044, opened 2022-06-15
It is not possible to transfer the prefix to the RedisLocking


pashtetus1:
In my redis cluster, rights are given strictly by prefixes. This makes work impossible. Although in other redis providers it is possible to transfer prefix to Lockproviders there is no such possibility.
```
    public static class ServiceCollectionExtensions
    {
        public static WorkflowOptions UseRedisQueues(this WorkflowOptions options, string connectionString, string prefix)
        {
            options.UseQueueProvider(sp => new RedisQueueProvider(connectionString, prefix, sp.GetService<ILoggerFactory>()));
            return options;
        }

        public static WorkflowOptions UseRedisLocking(this WorkflowOptions options, string connectionString /*, string prefix*/)
        {
            options.UseDistributedLockManager(sp => new RedisLockProvider(connectionString /*, prefix*/ sp.GetService<ILoggerFactory>()));
            return options;
        }

        public static WorkflowOptions UseRedisPersistence(this WorkflowOptions options, string connectionString, string prefix, bool deleteComplete = false)
        {
            options.UsePersistence(sp => new RedisPersistenceProvider(connectionString, prefix, deleteComplete, sp.GetService<ILoggerFactory>()));
            return options;
        }

        public static WorkflowOptions UseRedisEventHub(this WorkflowOptions options, string connectionString, string channel)
        {
            options.UseEventHub(sp => new RedisLifeCycleEventHub(connectionString, channel, sp.GetService<ILoggerFactory>()));
            return options;
        }
    }
```

PR #1059, opened 2022-06-23
Add prefix to RedisLockProvider #1044

uuggene:
We use the same Redis connection string for all services which run WorkflowCore and have their own databases.
Current behavior locks other services from polling runnable workflows, events, etc. It uses on key to lock for different services.
Adding prefix can solve this problem.

And I also found that there's an open issue to have it.

### DIFF
--- a/src/providers/WorkflowCore.Providers.Redis/ServiceCollectionExtensions.cs
+++ b/src/providers/WorkflowCore.Providers.Redis/ServiceCollectionExtensions.cs
@@ -13,9 +13,9 @@ namespace Microsoft.Extensions.DependencyInjection
             return options;
         }
 
-        public static WorkflowOptions UseRedisLocking(this WorkflowOptions options, string connectionString)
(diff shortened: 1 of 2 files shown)


src/providers/WorkflowCore.Providers.Redis/ServiceCollectionExtensions.cs
```diff
@@ -13,9 +13,9 @@ public static WorkflowOptions UseRedisQueues(this WorkflowOptions options, strin
             return options;
         }
 
-        public static WorkflowOptions UseRedisLocking(this WorkflowOptions options, string connectionString)
+        public static WorkflowOptions UseRedisLocking(this WorkflowOptions options, string connectionString, string prefix)
```
danielgerlag, line 16: Can we make this backwards compatible?

uuggene: Done

PR #1122, opened 2022-12-08
(Bug) Updated definition loader to evaluate children inputs

SavinPopescu:
**Describe the change**
Updated the way of getting children of a JObject in order to evaluate expressions for inputs that have child elements.

Used the select token method in order to select the children then evaluate the type of the token before adding to the stack of JObjects.

**Tests**
Created a new step, DynamicDataStep that I have added into stored-dynamic-definition.json with a complex object in Inputs.

**Breaking change**
No, this is an added functionality on top. It is a bug fix because children items were not evaluated.

**Additional context**
We discovered the issue when working for a dynamic workflow solution for one of our clients.

### DIFF
--- a/src/WorkflowCore.DSL/Services/DefinitionLoader.cs
+++ b/src/WorkflowCore.DSL/Services/DefinitionLoader.cs
@@ -400,20 +400,20 @@ namespace WorkflowCore.Services.DefinitionStorage
 
                     foreach (var child in subobj.Children())
                     {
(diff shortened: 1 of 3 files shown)


EmanuelOprea:
Exactly what I was looking for!
I had an issue setting JSON with multiple child complex objects/arrays.

wfcore-manager:
Closing as stale (opened 2022). Please reopen if still relevant.

Issue #1128, opened 2022-12-20
Migration issue with .NET 7


ssougnez:
**Describe the bug**

I currently have an application running with .NET 5 and wfc 3.4.0. I've been working on updating its dependencies and everything went fine until .NET 7. Indeed, when upgrading to .NET 6 and wfc 3.8.1, everything went as expected. However, today, I tried the same upgrade but towards .NET 7 but the migration of wfc failed with the following error:

> System.AggregateException: 'One or more errors occurred. (Method not found: 'Microsoft.EntityFrameworkCore.Migrations.Operations.Builders.OperationBuilder`1<Microsoft.EntityFrameworkCore.Migrations.Operations.CreateIndexOperation> Microsoft.EntityFrameworkCore.Migrations.MigrationBuilder.CreateIndex(System.String, System.String, System.String[], System.String, Boolean, System.String)'.)'

To pint point the issue, I first tried to upgrade from ".NET 5/wfc 3.4.0" to ".NET 6/wfc 3.8.1" then to ".NET 7/wfc 3.8.1" and with this approach it worked. I guess that the failing migration could be applied with .NET 6 but not with .NET 7.

I'm OK to deploy the application in two steps to fix this issue but I'm wondering whether this issue will occur again in the future. What if you create your database migrations with an old version of EF Core generating code that is not compatible starting with .NET 7? I'm just shooting in the dark here, but I think this issue should be investigated to ensure that future updates won't break.

**To Reproduce**

- Use wfc 3.4.0 with .NET 5
- Upgrade to wfc 3.8.1 with .NET 7

**Expected behavior**

Migrations should be applied without any issue.

**Additional contact**

I also use:

- WorkflowCore.Persistence.SqlServer 3.8.1
- WorkflowCore.Providers.Redis 3.8.1

danielgerlag:
I don't think this is a .NET 7 issue... you should be able to use .NET 7 with EF Core 6....
It is an EF Core issue, the culture of the EF Core team is to make breaking changes from version to version.

ssougnez:
But if this is the case, it might mean that the SQL provider wouldn't work on a new project using .NET 7 because of the past migrations...

It's a weird one... I'll do the upgrade in two times but I feared that a future migration might fail because of this.

danielgerlag:
I regret having used EF core for several of the DB providers... I want to say we should start thinking about rewriting them without EF Core.

ssougnez:
[...]

PR #1168, opened 2023-05-17
Update EFCore to 7

elexisvenator:
**Describe the change**
Change all .net 6 package references to entity framework to 7.*.

**Tests**
All current tests pass, there shouldnt be a need for new tests.

**Breaking change**
This is NOT backwards compatible with EF core 6 or lower.

**Additional context**
This is a PR that will address #1128 , however there should be discussion about if this is the right approach, or if another way should be done to maintain compatibility with EFCore 6, such as creating a separate package.

### DIFF
--- a/src/providers/WorkflowCore.Persistence.EntityFramework/WorkflowCore.Persistence.EntityFramework.csproj
+++ b/src/providers/WorkflowCore.Persistence.EntityFramework/WorkflowCore.Persistence.EntityFramework.csproj
@@ -22,7 +22,7 @@
   </ItemGroup>
 
   <ItemGroup Condition=" '$(TargetFramework)' == 'net6.0' ">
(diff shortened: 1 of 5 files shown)


ErikEJ:
Why do you need this change?

elexisvenator:
Our application makes used of libraries that require Npgsql 7.0.0+. (The library in question is [Marten 6.0](https://martendb.io/)). The earliest version of EntityFramework Core that supports this version of Npgsql is also 7.0.0+. There are also breaking changes in EntityFramework Core that means manually pinning the versions to 7 do not work (see #1128).

ssougnez:
Up for this one?

I tried starting a new .NET 7 application with WorkflowCore and it does not work because of https://github.com/danielgerlag/workflow-core/issues/1128

danielgerlag:
Sorry for the delay. I will get this one through in the next day or two.

PR #1176, opened 2023-06-21
Implemented Event table purger

viktorshevchenko210:
**Describe the change**
Implemented Event table purger

Implemented EventsPurger in a similar way to WorkflowPurger.

**Tests**
Yes

**Breaking change**
No

**Additional context**
We had 16 million rows and it broke our PROD.

### DIFF
--- a/src/providers/WorkflowCore.Persistence.EntityFramework/Services/EventsPurger.cs
+++ b/src/providers/WorkflowCore.Persistence.EntityFramework/Services/EventsPurger.cs
@@ -1,0 +1,46 @@
+﻿using System;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using WorkflowCore.Interface;
+using WorkflowCore.Persistence.EntityFramework.Interfaces;
+using WorkflowCore.Persistence.EntityFramework.Models;
+
+namespace WorkflowCore.Persistence.EntityFramework.Services
+{
+    public class EventsPurger : IEventsPurger
+    {
+        private readonly IWorkflowDbContextFactory _contextFactory;
+
+        public EventsPurger(IWorkflowDbContextFactory contextFactory)
+        {
+            _contextFactory = contextFactory;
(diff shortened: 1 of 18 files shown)


viktorshevchenko210:
@danielgerlag  could you please review this PR? Event table is never cleared.

src/providers/WorkflowCore.Persistence.EntityFramework/Services/EventsPurger.cs
```diff
@@ -0,0 +1,46 @@
...
+        public async Task PurgeEvents(DateTime olderThan, CancellationToken cancellationToken = default)
+        {
+            var olderThanUtc = olderThan.ToUniversalTime();
+            using (var db = ConstructDbContext())
+            {
+                var events = db.Set<PersistedEvent>()
+                    .Where(x => x.EventTime < olderThanUtc &&
+                                x.IsProcessed == true);
+
+                db.RemoveRange(events);
```
SergiiKram, line 35: @viktorshevchenko210 Since the WorkflowCore uses EF Core 7 now, can we use EF Core 7 bulk delete feature here?

ErikEJ: This will most likely time out with the default timeout if 30 seconds.  Needs options for timeout in seconds and batch size to fit each specific usage.

starushykart: Mb it makes sense to delete Subscriptions related to this event?

viktorshevchenko210: @ErikEJ Added options for timeout and batch size.

SergiiKram: Thanks, Viktor!

Issue #1181, opened 2023-06-26
ActivityController.Token is private and is not open for extension


martikyan:
**Is your feature request related to a problem? Please describe.**
[WorkflowCore/Services/ActivityController.Token](https://github.com/danielgerlag/workflow-core/blob/12be88701780e1f1b2d5b5b5d98e7ce4df3e9913/src/WorkflowCore/Services/ActivityController.cs#L108) is private which makes `ActivityController` not open for extension.

**Describe the solution you'd like**
I would like to propose separating the `Token` from `ActivityController` and making it a public class.

danielgerlag:
The point is encapsulation, so that these things can be refactored in the future without breaking anything external.  Once you make it public, it becomes part of an API surface.

Issue #1180, opened 2023-06-26
EntityFramework/ExtensionMethods are private and are not open for extension


martikyan:
**Is your feature request related to a problem? Please describe.**
[ExtensionMethods](https://github.com/danielgerlag/workflow-core/blob/12be88701780e1f1b2d5b5b5d98e7ce4df3e9913/src/providers/WorkflowCore.Persistence.EntityFramework/ExtensionMethods.cs#L10C5-L10C5) in the EntityFramework provider is currently internal and is not open for extension.

**Describe the solution you'd like**
I would like to propose converting `ExtensionMethods` into a service with extensions made public.

**Describe alternatives you've considered**
One alternative would be to simply mark those methods as public

**More context**
 The service will most probably be dependent on `JsonSerializerSettings`. See also #1178

danielgerlag:
The point is encapsulation, so that these things can be refactored in the future without breaking anything external.  Once you make it public, it becomes part of an API surface.

Issue #1179, opened 2023-06-26
Increase the [MaxLength] for ExternalToken in PersistedSubscription in EntityFramework provider


martikyan:
**Is your feature request related to a problem? Please describe.**
If activities have encoded tokens longer than the current limit of 200 characters, the `ExternalToken` will not be persisted.

**Describe the solution you'd like**
I would like to suggest increasing the `[MaxLength]` attribute for `ExternalToken` to a more reasonable length. 500 seems to work fine.

Issue #1178, opened 2023-06-26
Provide JsonSerializerSettings customization for EntityFramework provider


martikyan:
**Is your feature request related to a problem? Please describe.**
There is a use case where we need specific `JsonSerializerSettings` for the EntityFramework provider. However, the settings are not exposed and, therefore, cannot be customized.

**Describe the solution you'd like**
I would like `WorkflowOptions` to have a method that allows the setting of custom fields for the serializer settings.

PR #1184, opened 2023-06-30
Services made more extensible, fixed exceptions on some use cases

martikyan:
**Describe the change**
- Increased subscription ExternalToken length limit
- Persistence JsonSerializerSettings exposed to WorkflowOptions
- Modified a few access modifiers to enable extensibility

**Tests**
I was not able to run all tests because they were failing on my machine.

**Breaking change**
No

**Additional context**
This PR is to be squash merged
#1178 
#1179 
#1180 
#1181


src/WorkflowCore/Models/WorkflowOptions.cs
```diff
@@ -26,6 +27,10 @@ public WorkflowOptions(IServiceCollection services)
             PollInterval = TimeSpan.FromSeconds(10);
             IdleTime = TimeSpan.FromMilliseconds(100);
             ErrorRetryInterval = TimeSpan.FromSeconds(60);
+            JsonSerializerSettings = new JsonSerializerSettings
```
danielgerlag, line 30: I think this belongs on the settings for a specific persistence provider that requires JSON, rather than on the global workflow settings

src/providers/WorkflowCore.Persistence.EntityFramework/Models/PersistedSubscription.cs
```diff
@@ -30,7 +30,7 @@ public class PersistedSubscription
 
         public string SubscriptionData { get; set; }
         
-        [MaxLength(200)]
+        [MaxLength(500)]
```
danielgerlag, line 33: We need to generate migrations for all SQL Server, Postgres and MySQL when we change schema

src/providers/WorkflowCore.Persistence.EntityFramework/Services/EntityFrameworkPersistenceProvider.cs
```diff
@@ -16,12 +16,14 @@ public class EntityFrameworkPersistenceProvider : IPersistenceProvider
         private readonly bool _canCreateDB;
         private readonly bool _canMigrateDB;
         private readonly IWorkflowDbContextFactory _contextFactory;
+        private readonly ModelConverterService _modelConverterService;
 
         public bool SupportsScheduledCommands => true;
 
-        public EntityFrameworkPersistenceProvider(IWorkflowDbContextFactory contextFactory, bool canCreateDB, bool canMigrateDB)
+        public EntityFrameworkPersistenceProvider(IWorkflowDbContextFactory contextFactory, ModelConverterService modelConverterService, bool canCreateDB, bool canMigrateDB)
```
danielgerlag, line 23: We should depend on abstractions rather than concrete implementations, this should be an interface

src/WorkflowCore/Models/Token.cs
```diff
@@ -0,0 +1,37 @@
+using System;
+using System.Linq;
+using System.Text;
+using Newtonsoft.Json;
+
+namespace WorkflowCore.Models
+{
+    public class Token
```
danielgerlag, line 8: Why does this need to be public?

danielgerlag (changes requested):

martikyan:
@danielgerlag sorry, closing this PR

PR #1187, opened 2023-07-11
RedisProvider - Remove the actual event record if we're tidying up after ourselves.

ChrisBertrand:
**Describe the change**
Following the _removeComplete logic on PersistWorkflow, also remove events if this is set.

Followed the same logic from the PersistWorkflow method.

**Tests**
Tested locally.

**Breaking change**
No. There are no changes to the method signature or interface.  

**Additional context**
This is really useful in the Redis provider as there is no cleanup functionality available, finding the right keys within the events hash key is rather difficult, having it as part of the main logic would simplify this.


ChrisBertrand:
@danielgerlag Would be great to get feedback on this when you're next free. Thanks.

wfcore-manager:
Closing as stale (no updates since 2023). Please reopen if still relevant.

PR #1198, opened 2023-08-06
Fixed usage of null WorkflowStep.Name in WorkflowActivity diagnostic service.

jakenuts:
**Describe the change**
In WorkflowActivity.Enrich(WorkflowStep workflowStep) there was a check for a null or empty WorkflowStep.Name and "inline" was used in it's place in the activity display name. A line immediately after referenced WorkflowStep.Name without that check and causes a NullReferenceException if System.Diagnostics.ActivitySource & System.Diagnostics.Activity are used by the host or workflow implementation.

Replaced workflowStep.Name with stepName defined several lines above which either contains the name or "inline" for unnamed steps (Sagas, Snline, etc).

**Tests**
Did you cover your changes with tests? No.

**Breaking change**
Do you changes break compatibility with previous versions? No.

**Additional context**
As far as I can tell usage of ActivitySource and related services like OpenTelemetry are not possible while this bug remains.


jakenuts:
Added additional nullable check for Body.Type.Name as that also is possibly null (End() step)

Issue #1155, opened 2023-03-21
NullReferenceException after upgrading 3.4.0 -> 3.8.1


R4cOOn:
We upgraded our library from 3.4.0 to 3.8.1.  
We use PostgreSQL 14 as our backend and had to do a "manual upgrade" of the DB because the EF version used by the library wasn't supported by the the version we had.

Now things work most of the time but we have cases where interacting with the workflows raises `NullReferenceException`s.  
Looking at the code it happens in `WorkflowActivity.Enrich(WorkflowStep workflowStep)` at line 65:
```
      current.SetTag("workflow.step.type", (object) workflowStep.BodyType.Name);
```
our `workflowStep.BodyType` is `null` because we use `StepBodyAsync` (as described in https://workflow-core.readthedocs.io/en/stable/getting-started/) and not the `WorkflowStep` class.  
We also see log entries like `[11:17:09 WorkflowCore.Services.WorkflowExecutor DBG] Starting step null on workflow eca3fed6-42bf-4ec8-b21b-72261e995ed7` for the same reason.

I didn't see any release notes mentioning a breaking change.  
Is this a bug or do we need to make some changes?

We also see some `DbUpdateException` because of `ScheduledCommand` unique index constraint failing.  
Are these also expected?

Cheers.

qrzychu:
@danielgerlag  any news on that one?

still throwing. Fix is easy - just put a question mark there.

Any idea why this happens? How should we write our steps to avoid this problem?

We inherit the StepBody or StepBodyAsync

danielgerlag:
Thanks for discovering that.  We should not set the BodyType to null in the EndStep, that would be the correct solution.

DavidPx:
Does anyone have any workarounds for this?  Our build system needs the System.Linq.Dynamic.Core bump in v3.10 but our workflow no longer works after v3.6.2, when the open telemetry stuff was put in.

laurentiustamate94:
seems like with the latest release ([v3.11.0](https://github.com/danielgerlag/workflow-core/releases/tag/v3.11.0)), specifically #1281 this issue disappeared.

PR #1212, opened 2023-11-07
#1155: coalesce null

qrzychu:
**Describe the change**
this line throws an errors as logged in #1155


danielgerlag (changes requested):
Let's rather fix the EndStep

qrzychu:
what should be the body type then?

danielgerlag:
> what should be the body type then?

typeof(EndStep).Name

PR #1215, opened 2023-11-24
Upgrade to net8 and EF8

wallyrion:
**Describe the change**
Upgrade target framework from net6 to net 8
Upgrade EF7 to EF8 for net8 framework

**Tests**
All current tests pass, there shouldn't be a need for new tests. Only an additional target framework net8 added

**Breaking change**
This is NOT backwards compatible with EF core 7 or lower.

**Additional context**


wallyrion:
Looks like AppVeyor pipeline needs to be updated

knoxi:
+1 for .net 8

danielgerlag (approved):

Issue #1231, opened 2024-01-27
[Proposal] Add custom IServiceScope support so that Step only accesses service instances of this scope


ling921:
**Is your feature request related to a problem? Please describe.**

I have a scoped service `IMyService`. I have now obtained an `IMyService` instance. I hope that all steps will obtain this instance when the workflow is running. In this way, I can prevent the services between multiple workflows from interfering with each other.

**Describe the solution you'd like**

Add a method to `IWorkflowController` called `StartWorkflowWithScope`, which accept an additional parameter `IServiceScope`, and then save it in `WorkflowInstance`

**Describe alternatives you've considered**

None

**Additional context**

I would like to make a pull request for this.

vladimir-kovalyuk:
From my perspective what you would like to achieve is not guaranteed within the resilient long running process. The host may execute the half of steps then get restarted and continue executing the rest of steps. So you can not rely on the WFC that it would provide the same IMyService for all steps.
You may inject IServiceScope and IServiceScopeFactory into your step. Maybe these interfaces being injected would help you with your needs?

Issue #846, opened 2021-06-16
Scoped Service Dependencies Not Re-used


DavidPx:
**Describe the bug**
I have a workflow hosted in a .net API.  I have a data access layer dependency used by multiple steps which I'd like created once per request so I've configured with `AddScoped` in the service collection.  My steps are all registered with `AddTransient`.

However each time a step is created the shared dependency is also re-created, counter to the documented AddScoped behavior.   

**To Reproduce**
I've modified the Hello World sample to demonstrate the problem.
https://github.com/DavidPx/workflow-core/tree/scopedepdencybug/src/samples/WorkflowCore.Sample01/Steps

When you run the program you can see how each step gets a new instance:

```
Shared Dependency Id = 794023b7-0f1d-42f9-8d46-2cdf16e806ec
Hello world
Shared Dependency Id = 91e5a41c-3a6e-4f97-927e-4099a96a8420
Goodbye world
```

**Expected behavior**
I would expect that each step would receive the same instance of the scoped dependency:
```
Shared Dependency Id = 91e5a41c-3a6e-4f97-927e-4099a96a8420
Hello world
Shared Dependency Id = 91e5a41c-3a6e-4f97-927e-4099a96a8420
Goodbye world
```

Locally I threw together a PoC program with two Transient services that depend upon a Scoped service, and the lifetimes are managed as I would expect.

**Additional context**
Here is my little PoC API..

https://github.com/DavidPx/DiTest

The two matching guids are from the scoped/shared depenencey that the two transient services depend upon.

```
{
  "transientId": "6cf3e26e-a32f-48d8-aeab-26467f5c7e86",
  "secondTransientId": "d65ead19-e7ce-4da5-b1d9-ca6592dd8b10",
  "transientDependencyId": "efb0efdd-22ba-4682-93c5-4a42763f36dc",
  "transient2DependencyId": "efb0efdd-22ba-4682-93c5-4a42763f36dc",
  "singletonService": "ec454a98-86af-4dfb-b473-bc9ef2db3da9"
}
```

DavidPx:
I debugged the framework a bit and think that behavior is caused by the new scope create per step:

https://github.com/danielgerlag/workflow-core/blob/master/src/WorkflowCore/Services/WorkflowExecutor.cs#L158

I don't see a way to configure the scope provider to use one scope per request... could I supply one?

danielgerlag:
[...]

PR #1232, opened 2024-01-27
[Feature] Add scope support for workflow

ling921:
**Describe the change**

This pull request is to make `Step` to only obtain services from a specific scope. 

Related issues: #1231 #846 

Add a method to `IWorkflowController` called `StartWorkflowWithScope`, which accept an additional parameter `IServiceScope`, and then save it in `WorkflowInstance`

**Tests**

Yes

**Breaking change**

No

**Additional context**

None

### DIFF
--- a/WorkflowCore.sln
+++ b/WorkflowCore.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 16
-VisualStudioVersion = 16.0.29509.3
+# Visual Studio Version 17
+VisualStudioVersion = 17.9.34518.117
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "src", "src", "{EF47161E-E399-451C-BDE8-E92AAD3BD761}"
 EndProject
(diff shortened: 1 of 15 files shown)


danielgerlag:
@ling921 thanks for the effort.  Unfortunately, this does not make sense to wrap an entire workflow in a scope.  When the workflow is executed, the entire thing "may" be executed all at once, or it may be partially executed and continue several days later, or on another machine. So we cannot preserve the scope across these time and node boundaries.
We do wrap step executions within a scope [here](https://github.com/danielgerlag/workflow-core/blob/master/src/WorkflowCore/Services/WorkflowExecutor.cs#L159)

ling921:
I understand this, and I haven't changed its default behavior. If users expect wrokflow steps to be executed within a certain scope, they should manage the creation and disposal of the scope themselves.

Also, I think the design of this scope is meaningful. I can define a scoped service to track or change the step's execution, and even share some unmanaged resources, etc. All of these are more suitable to be written in a service rather than within `TData`.

The primary challenge is ensuring the effective management of the scope, especially regarding proper disposal. In my current design, it is imperative for the user to proficiently handle scope disposal. With correct scope disposal, I believe there should be no issues with long-term retention.

There may be a more optimal design that can achieve this.

PR #1286, opened 2024-08-14
Feat update rabbitmq

JoaquimInGit:
**Describe the change**
Update the RabbitMq Client library to the latest stable [6.8.1](https://github.com/rabbitmq/rabbitmq-dotnet-client/releases/tag/v6.8.1)

The only breaking change that regard workflow core is consumer delivery payloads are now [ReadOnlyMemory<byte>](https://learn.microsoft.com/en-us/dotnet/api/system.readonlymemory-1?view=net-8.0) instead of byte[]
in [.netstandart 2.1](https://learn.microsoft.com/en-us/dotnet/api/system.text.encoding.getstring?view=netstandard-2.1#System_Text_Encoding_GetString_System_ReadOnlySpan_System_Byte__) we could do .Span()
in [.netstandart 2.0](https://learn.microsoft.com/en-us/dotnet/api/system.text.encoding.getstring?view=netstandard-2.0#System_Text_Encoding_GetString_System_ReadOnlySpan_System_Byte__) .ToArray() does the job.

**Tests**
Tests already existed 

**Breaking change**
Versions with rabbit <6.0.0 will not work

**Additional context**

Any chance this could be added, we are in a pinch :D
Thanks for the work <3

### DIFF
--- a/src/providers/WorkflowCore.QueueProviders.RabbitMQ/Services/RabbitMQProvider.cs
+++ b/src/providers/WorkflowCore.QueueProviders.RabbitMQ/Services/RabbitMQProvider.cs
@@ -63,7 +63,7 @@ namespace WorkflowCore.QueueProviders.RabbitMQ.Services
                 var msg = channel.BasicGet(_queueNameProvider.GetQueueName(queue), false);
                 if (msg != null)
                 {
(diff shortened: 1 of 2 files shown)


pedrojreis:
+1 for up to date RabbitMQ!

PR #1311, opened 2024-10-23
Upgrade to Mongo Version 2.30

dsbegnoche:
Upgrading mongo to version 2.30. MongoDb started strong naming their packages with plenty of pushback. However it appears they are committed to this change, and therefore needing all projects to upgrade to 2.28+.  A major version 3.0 exists, but would require more testing than just this upgrade and ensure builds are passing.
Discussion on the change:
https://www.mongodb.com/community/forums/t/net-c-driver-strong-naming/291649/16


gmiserez:
Hi @danielgerlag @glucaci, 
Did you have the opportunity to review this topic? Not going to lie, my team is eagerly awaiting this update (and the future update to mongodb 3.0.0), as we have 6 different libraries with dependencies to mongodb, and need to keep the dependencies aligned.

danielgerlag (approved):

PR #1341, opened 2025-03-10
Feat/pre middelware syncrunner

zewa666:
**Describe the change**
The SyncWorfklowRunner did not call any potentially registered PreWorkflowMiddleware. This fixes it by adding the same call as the WorkflowController

Replicate use-case from WorkflowController

**Tests**
Yup, added a new test that verifies the call of `RunPreMiddleware`

**Breaking change**
Nope

**Additional context**
Let me know if anything else is needed.

### DIFF
--- a/src/WorkflowCore/Services/SyncWorkflowRunner.cs
+++ b/src/WorkflowCore/Services/SyncWorkflowRunner.cs
@@ -17,8 +17,9 @@ namespace WorkflowCore.Services
         private readonly IExecutionPointerFactory _pointerFactory;
         private readonly IQueueProvider _queueService;
         private readonly IDateTimeProvider _dateTimeProvider;
+        private readonly IWorkflowMiddlewareRunner _middlewareRunner;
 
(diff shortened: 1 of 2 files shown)


zewa666:
@danielgerlag any chance this gets a quick review?

danielgerlag (approved):

zewa666:
@glucaci any chance to get a review from you and this PR merged?

danielgerlag:
## Review & Test Fix

I reviewed this PR and the production code change is correct  `SyncWorkflowRunner` was indeed missing the `RunPreMiddleware` call that `WorkflowController` already had.

However, the unit test `should_run_pre_middlewares` had two issues causing it to fail:

1. **1ms timeout too short**: `TimeSpan.FromMilliseconds(1)` caused `TaskCanceledException` before execution completed
2. **Workflow never completed**: The fake step used `ExecutionPipelineDirective.Next` in `InitForExecution`, but the mocked `IExecutionResultProcessor` was a no-op  meaning the execution pointer was never finalized, the workflow never transitioned to `Complete`, and the execute loop ran until timeout

**Fix applied** (pushed to `feat/pre-middelware-syncrunner`):
- Changed `InitForExecution` to return `ExecutionPipelineDirective.EndWorkflow` so the executor immediately completes the workflow (we only need to verify pre-middleware gets called, not full step execution)
- Increased timeout to 5 seconds as a safety net

**Test results after fix**: 59/59 unit tests pass, 44/44 integration tests pass.

PR #1438, opened 2026-09-04
Added null-handling when recording host startup exception to current …

cervengoc:
**Describe the change**
In WorkflowHost StartAsync if any exception occures, and there is no ActivityListener for WorkflowCore, then the catch block throws NullReferenceException due to accessing the null activity. This hides the original startup exception and makes impossible to see what had actually happen (eg. RabbitMQ queue provider connection issue, etc.)

Added a null-conditional operator to AddException

**Tests**
No

**Breaking change**
No


danielgerlag (approved):
LGTM. StartHost() can return null when ActivitySource has no listeners (StartActivity contract); finally already used activity?.Dispose(), catch did not — this was masking real startup failures behind an NRE. One-line fix is correct.

Issue #1417, opened 2026-03-04
QueueConsumer _secondPasses are not cleared when Workflow has ended or terminated.


stechklvana:
**Describe the bug**
QueueConsumer _secondPasses are not cleared when Workflow has ended or terminated.

**To Reproduce**
We used WorkflowPurger to clean Workflows with Status 2 or 3 to clean SQL Server database. After that 2 errors appeared for the deleted records:

"_Error processing workflow "8fa9ba29-2232-4705-b8bd-2833a56147cc"__"
"_Error executing item 8fa9ba29-2232-4705-b8bd-2833a56147cc - Sequence contains no elements._"

That workflowId (InstanceId) was one of those which were deleted from the database. We used SingleNodeQueueProvider but I dont think this will change with different provider.

**Expected behavior**
After Workflow End (Status=2) or Termination (Status=3) the workflowId should be cleared from all queues.

**Additional context**
Maybe you only need to add cleaning _secondPasses to the finally block of the ExecuteItem method. Something like this:

` finally
 {
     waitHandle.Set();
     _secondPasses.TryRemove(itemId); // Fix #1: cleanup i when exception/cancellation occurs
     lock (_activeTasks)
     {
         _activeTasks.Remove(itemId);
     }
     lock (_runningTasksLock)
     {
         _runningTasks.RemoveAll(t => t.IsCompleted); // Fix #2: cleanup done tasks
     }
 }`

wfcore-manager:
Opened https://github.com/danielgerlag/workflow-core/pull/1444 for this: clear `_secondPasses` in `QueueConsumer.ExecuteItem` `finally`, and null-guard after `GetWorkflowInstance` in `WorkflowConsumer` / `IndexConsumer` when Redis (or another store) has already deleted the instance. No change to `deleteCompleted` semantics.

PR #1154, opened 2023-03-17
fix issue #1029 EnableIndexes is set to false but Index-related work are still being queued

brucezlata:
**Describe the change**
queue and dequeu by check WorkflowOptions, EnableIndexes etc.

**Describe your implementation or design**
WorkflowOptions options in RedisQueueProvider Constuction

**Tests**
Did you cover your changes with tests?

**Breaking change**
Do you changes break compatibility with previous versions?

**Additional context**
Any additional information you'd like to provide?

PR #1446, opened 2026-09-08
Harden Redis queue uniqueness with atomic Lua enqueue

danielgerlag:
**Describe the change**
`RedisQueueProvider.QueueWork` de-duped pending ids with a non-atomic `LINSERT` + conditional `RPUSH` / `LREM`. Two concurrent enqueues of a missing id could both miss, both `RPUSH`, and leave duplicate pending LIST entries. Dequeue was (and remains) `LPOP`.

This change makes pending-id uniqueness atomic **inside the Redis queue provider only**. `IQueueProvider` is unchanged. Issue #1417 / PR #1444 (`QueueConsumer._secondPasses` after purge) are a separate consumer-side problem and are not addressed here.

Smallest upgrade-safe design: keep the existing LIST keys and the existing LINSERT/RPUSH/LREM algorithm, but run that sequence in one Lua `EVAL` so Redis serializes it.

Rejected alternative (SADD membership + RPUSH, LPOP+SREM): extra key, rolling-upgrade skew between old processes (LIST-only) and new processes (LIST+SET), and a crash window unless dequeue is also a single Lua. Safer for existing deployments is “no second key.”

`DequeueWork` stays a single `LPOP`. Because there is no companion SET, there is nothing to desync if a process dies after pop.

**Tests**
Added `RedisQueueProviderFixture` in `WorkflowCore.Tests.Redis` (existing Squadron Redis collection — this repo does not use Testcontainers for Redis):

- Concurrent `QueueWork` of a missing id leaves exactly one pending entry
- Re-`QueueWork` while pending is a no-op and preserves FIFO position
- Concurrent dequeue of unique ids returns each id once and leaves the LIST empty
- Dequeue then re-enqueue of the same id succeeds
- Pre-existing duplicate LIST entries are not grown by `QueueWork` and drain via `LPOP`

**Breaking change**
No required migration or flush. Same Redis keys, same intended uniqueness semantics, now actually atomic.

**Compatibility**

This design was chosen so existing Redis keys keep working. Details:

- **Existing LIST keys that already contain duplicate entries:** No migration. Duplicates stay until they are dequeued (`LPOP` drains them one by one). A new `QueueWork` of that id does **not** add another occurrence. Operators do **not** need a one-time flush; optionally deleting `{prefix}-workflows|events|index` would only drop in-flight work (not recommended).
- **Rolling upgrades (mixed old/new `RedisQueueProvider` processes):** Both generations share the same LIST. New processes enqueue atomically among themselves. An old process can still race its own multi-command sequence (or interleave `LINSERT` miss + `RPUSH` next to a new process) and insert a duplicate until every node is upgraded. After the rollout completes, uniqueness holds.
- **Extra Redis keys:** None. No companion SET, so no naming/TTL/cleanup and no SET/LIST drift.
- **Crash window between pop and membership remove:** None. Dequeue is one `LPOP`. An id cannot get stuck in a membership set or be lost because of a failed follow-up `SREM`.
- **Semantic change (re-QueueWork while pending):** Stronger no-op. The old code *intended* not to add a second pending entry, but concurrent misses could both `RPUSH`. After this change, re-queue of an id that is still pending is always a no-op. After dequeue, the same id may be queued again. FIFO position of an already-pending id is unchanged (it is not moved to the tail).
- **One-time migration/flush:** Not required.
- **Redis version:** Script uses `LINSERT` (2.2+) and Lua (2.6+). No `LPOS` / Redis 6 requirement.
- **`IQueueProvider` / other providers:** Unchanged. Production behavior outside `RedisQueueProvider` is unchanged.

**Additional context**
Queue key format is unchanged: `{prefix}-workflows`, `{prefix}-events`, `{prefix}-index`.

<div><a href="https://cursor.com/agents/bc-747bff44-fdb4-4802-a451-0449ad52891d?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-747bff44-fdb4-4802-a451-0449ad52891d&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/src/providers/WorkflowCore.Providers.Redis/Services/RedisQueueProvider.cs
+++ b/src/providers/WorkflowCore.Providers.Redis/Services/RedisQueueProvider.cs
@@ -8,8 +8,41 @@ using WorkflowCore.Interface;
 
 namespace WorkflowCore.Providers.Redis.Services
 {
+    /// <summary>
+    /// Redis LIST-backed <see cref="IQueueProvider"/>. Pending ids are unique per queue.
+    /// </summary>
+    /// <remarks>
+    /// Compatibility:
+    /// <list type="bullet">
+    /// <item>Uses the existing LIST keys only ({prefix}-workflows|events|index). No companion SET,
+    /// so there is no extra key to name, TTL, or clean up, and no pop/SREM crash window.</item>
+    /// <item>Existing LIST keys (including those that already contain duplicate entries) need no
+    /// migration or flush. Pre-existing duplicates drain via <see cref="DequeueWork"/>; a new
+    /// enqueue of that id does not add another occurrence.</item>
+    /// <item>Rolling upgrades with mixed old/new processes share the same LIST. New processes
+    /// enqueue atomically among themselves; an old process can still race its own
+    /// LINSERT/RPUSH sequence and insert a duplicate until every node is upgraded.</item>
+    /// <item>Re-QueueWork while an id is still pending is a no-op (the previous multi-command
+    /// de-dupe intended this, but concurrent misses could both RPUSH). After dequeue, the id
+    /// may be queued again.</item>
+    /// </list>
+    /// </remarks>
     public class RedisQueueProvider : IQueueProvider
     {
+        /// <summary>
+        /// Atomic unique enqueue: same LINSERT / RPUSH / LREM sequence as before, but one EVAL
+        /// so two concurrent misses cannot both RPUSH. LINSERT exists on Redis 2.2+; Lua on 2.6+.
+        /// </summary>
+        private const string UniqueEnqueueScript = @"
+local n = redis.call('LINSERT', KEYS[1], 'BEFORE', ARGV[1], ARGV[1])
+if n == -1 or n == 0 then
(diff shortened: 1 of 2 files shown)


wfcore-manager:
Superseded by https://github.com/danielgerlag/workflow-core/pull/1447 — the LIST Lua uniqueness script is folded in there (default `RedisQueueStorage.List`), with selectable SortedSet, docs, and tests that compile against current `master` after #1154. Closing this PR to avoid a duplicate merge path.

